PR #9, opened 2015-05-23
fix build on mac

santazhang:
* `typedef unsigned int uint`
  `uint` isn't a standard type - `unsigned int` is.

* add missing `#include <string>`
  clang on mac complains about this

### DIFF
--- a/src/lib/sqllib.h
+++ b/src/lib/sqllib.h
@@ -1,6 +1,7 @@
 #ifndef __SQLLIB_H__
 #define __SQLLIB_H__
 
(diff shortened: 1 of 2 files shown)


torpedro:
Thanks!
Can you also post the clang flags (or Makefile) you used to compile the project? I'd like to include these settings in the Makefile for users to be able to choose their compiler.

santazhang:
I'm directly using the project Makefile. There's no need to change the compiler flags.

torpedro:
Ok thank!

PR #17, opened 2016-04-30
Use C++ standard int types to allow compilation with clang++ also

ahalam:
`uint` is only available on Linux. Other than this, there seems to be no other dependency on the compiler. Replacing `unit` with `uintmax_t` (available since c++11 standard, ref: http://en.cppreference.com/w/cpp/types/integer)


torpedro:
Thanks, switching to standard types is definitely a good idea!

PR #20, opened 2016-10-05
Memory leaks and corruption fixes

ahalam:
- Added debug symbols for ease of development
- Switched from new/delete <-> malloc/free to avoid mismatch. Tool used AddressSanitizer
- Memory leak fixes. Using Valgrind.


torpedro:
Thanks for your work!
We used it as an inspiration to use automatic checks for memory leaks during CI builds and fixed the existing leaks based on that in PR #24.

PR #45, opened 2017-06-12
Fix: use correct format specifiers for int64

timzimmermann:
This PR fixes two compiler warnings regarding format specifiers (macOS / clang):

```c++
src/util/sqlhelper.cpp:14:52: warning:
      format specifies type 'long' but the argument has type 'int64_t' (aka 'long long') [-Wformat]
    printf("%s%ld  \n", indent(numIndent).c_str(), val);
              ~~~                                  ^~~
              %lld
sql-parser/src/util/sqlhelper.cpp:29:50: warning:
      format specifies type 'unsigned long' but the argument has type 'uint64_t' (aka 'unsigned long long') [-Wformat]
    printf("%s%lu\n", indent(numIndent).c_str(), val);
              ~~~                                ^~~
              %llu

### DIFF
--- a/example/example.cpp
+++ b/example/example.cpp
@@ -25,7 +25,7 @@ int main(int argc, char* argv[]) {
     printf("Parsed successfully!\n");
     printf("Number of statements: %lu\n", result.size());
 
-    for (uint i = 0; i < result.size(); ++i) {
+    for (auto i = 0u; i < result.size(); ++i) {
       // Print a statement summary.
       hsql::printStatementInfo(result.getStatement(i));
     }
--- a/src/util/sqlhelper.cpp
+++ b/src/util/sqlhelper.cpp
@@ -1,6 +1,6 @@
 
 #include "sqlhelper.h"
-#include <stdio.h>
+#include <iostream>
 #include <string>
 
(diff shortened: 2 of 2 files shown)


torpedro:
Thanks :+1: 
I'm guessing this is a specific distribution/compiler issue.
This is not occurring on Ubuntu with g++ 6.3.0, clang 3.8 or clang 4.0.
It's not occurring on our Jenkins CI builds either.
So we'll need a different solution working on both platforms.

This is mentioning a MacOS vs. Ubuntu issue: https://stackoverflow.com/questions/14073877/printf-and-llu-vs-lu-on-os-x

timzimmermann:
As suggested in the link you provided, is there any reason to use `printf` instead of `cout`?

torpedro:
No particular reason, we can change the implementation to use `cout` instead.

timzimmermann:
Okay cool, thanks for looking into it. I changed all `[f]printf` calls in `sqlhelper.cpp`.
I also got an error when compiling `example`, which I fixed as well.

torpedro:
Looks good! Thanks!

PR #46, opened 2017-06-12
Fix formatter warnings

TwoLaid:
This PR fixes the following warnings on current `g++` and `clang` compilers:

```C++
src/util/sqlhelper.cpp:14:52: warning: format specifies type 'long' but the argument has type 'int64_t' (aka 'long long') [-Wformat]
    printf("%s%ld  \n", indent(numIndent).c_str(), val);
              ~~~                                  ^~~
              %lld
src/util/sqlhelper.cpp:29:50: warning: format specifies type 'unsigned long' but the argument has type 'uint64_t' (aka 'unsigned long long') [-Wformat]
    printf("%s%lu\n", indent(numIndent).c_str(), val);
              ~~~                                ^~~
              %llu
```

### DIFF
--- a/src/util/sqlhelper.cpp
+++ b/src/util/sqlhelper.cpp
@@ -11,7 +11,7 @@ namespace hsql {
     return std::string(numIndent, '\t');
   }
   void inprint(int64_t val, uintmax_t numIndent) {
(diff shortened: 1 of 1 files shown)


torpedro:
Thanks :+1:
See #45, which addresses the same issue but was opened a few moments earlier.

PR #48, opened 2017-06-22
Add enum types for formerly 'simple' operands

timzimmermann:
This PR splits up `kOpSimple`, which was being used to describe operators that can be represented as a single char. The advantage of splitting those up is that we can more easily differentiate between them and do not need to compare both the type and the character.

### DIFF
--- a/src/parser/bison_parser.y
+++ b/src/parser/bison_parser.y
@@ -689,12 +689,12 @@ unary_expr:
 
 binary_expr:
 		comp_expr
-	|	operand '-' operand			{ $$ = Expr::makeOpBinary($1, '-', $3); }
-	|	operand '+' operand			{ $$ = Expr::makeOpBinary($1, '+', $3); }
-	|	operand '/' operand			{ $$ = Expr::makeOpBinary($1, '/', $3); }
-	|	operand '*' operand			{ $$ = Expr::makeOpBinary($1, '*', $3); }
-	|	operand '%' operand			{ $$ = Expr::makeOpBinary($1, '%', $3); }
-	|	operand '^' operand			{ $$ = Expr::makeOpBinary($1, '^', $3); }
+	|	operand '-' operand			{ $$ = Expr::makeOpBinary($1, kOpSubtract, $3); }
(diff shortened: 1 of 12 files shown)


src/parser/bison_parser.y
```diff
@@ -689,12 +689,12 @@ unary_expr:
...
-	|	operand '/' operand			{ $$ = Expr::makeOpBinary($1, '/', $3); }
-	|	operand '*' operand			{ $$ = Expr::makeOpBinary($1, '*', $3); }
-	|	operand '%' operand			{ $$ = Expr::makeOpBinary($1, '%', $3); }
-	|	operand '^' operand			{ $$ = Expr::makeOpBinary($1, '^', $3); }
+	|	operand '-' operand			{ $$ = Expr::makeOpBinary($1, kOpSubtract, $3); }
+	|	operand '+' operand			{ $$ = Expr::makeOpBinary($1, kOpAdd, $3); }
+	|	operand '/' operand			{ $$ = Expr::makeOpBinary($1, kOpDivide, $3); }
+	|	operand '*' operand			{ $$ = Expr::makeOpBinary($1, kOpMultiply, $3); }
+	|	operand '%' operand			{ $$ = Expr::makeOpBinary($1, kOpModulo, $3); }
+	|	operand '^' operand			{ $$ = Expr::makeOpBinary($1, kOpPower, $3); }
```
torpedro: The operator '^' presents a problem.
In the [PostgreSQL definition](https://www.postgresql.org/docs/7.4/static/functions-math.html) this is an exponentiation operator.
In [this definition](https://www.w3schools.com/sql/sql_operators.asp) they specify the operator as bitwise XOR.

Maybe we can change the names of all these operators to reference the symbol, instead of the semantics. This allows every user to decide themselves how they implement the operator.
So: kOpPower -> kOpCaret, kOpAdd -> kOpPlus, etc.

timzimmermann: Yes, Sven and I were discussing an appropriate naming scheme as well.
We'll change that.

PR #49, opened 2017-06-30
Many updates to grammar including support for arrays, ilike, natural joins with no predicates...

alkim0:
I made these updates because I needed to use them to parse my workload. The updates are not functional, but they do allow you to parse a greater variety of queries. Hopefully it is useful for you. 

List of changes:
- Added ISNULL, IS NULL, and IS NOT NULL to grammar
- Allow function calls with no arguments
- Added support for arrays
-- Can now create arrays in SQL queries with: `array[...]`
-- Indexed into arrays with: `myArray[0]`
- Support for CASE WHEN if there is no ELSE clause
- Allow joins between table_ref_atomic clauses
- Added the NULL literal
- Added ILIKE, CONCAT, and "!=" token for NOTEQUALS
- Added support for natural joins without without explicit join conditions
- Added optional ALL after set operators
- Added support for `table.*` to specify all columns in the table

When merging, you may want to squash a few commits since I didn't run astyle on the files until the very last commit.


torpedro:
Thanks, this is good work!
Before we can merge this back into the master, it's necessary to get the tests to succeed. Especially the conflict check for the grammar.
```
Checking for conflicts in the grammer...
bison_parser.y: warning: 55 shift/reduce conflicts [-Wconflicts-sr]
bison_parser.y: warning: 56 reduce/reduce conflicts [-Wconflicts-rr]
bison_parser.y:804.17-51: warning: rule useless in parser due to conflicts [-Wother]
make[1]: *** [test] Error 1
Conflict check failed!
```
If this does not work, we can also try to cherry-pick the features over time.

alkim0:
Ok, I've fixed the grammar up to make sure all the tests passed and that there are no grammar errors. However, it's been a while since my compilers class, so you may want to double check the precedence for the joins.

PR #53, opened 2017-07-17
Fix bug in print kExprFunctionRef expression.

NanXiao:
The [Expr::makeFunctionRef](https://github.com/hyrise/sql-parser/blob/fe35651bb5de1e7ca59d22c11359a149a45f19d3/src/sql/Expr.cpp#L108) uses `exprList` not `expr` to store expressions:  

```
Expr* Expr::makeFunctionRef(char* func_name, std::vector<Expr*>* exprList, bool distinct) {
    Expr* e = new Expr(kExprFunctionRef);
    e->name = func_name;
    e->exprList = exprList;
    e->distinct = distinct;
    return e;
  }
```

So during print [kExprFunctionRef expression](https://github.com/hyrise/sql-parser/blob/fe35651bb5de1e7ca59d22c11359a149a45f19d3/src/util/sqlhelper.cpp#L101):  

```
    case kExprFunctionRef:
      inprint(expr->name, numIndent);
      inprint(expr->expr->name, numIndent + 1);
      break;
```

`expr->expr` is a `nullptr`, and will cause program crash.

### DIFF
--- a/src/util/sqlhelper.cpp
+++ b/src/util/sqlhelper.cpp
@@ -100,7 +100,7 @@ namespace hsql {
       break;
     case kExprFunctionRef:
       inprint(expr->name, numIndent);
-      inprint(expr->expr->name, numIndent + 1);
+      for (Expr* e : *expr->exprList) inprint(e->name, numIndent + 1);
       break;
     case kExprOperator:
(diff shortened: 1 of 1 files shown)


torpedro:
Looks good, thank!

PR #58, opened 2017-08-01
Init expr

NanXiao:
The Expr struct need to be initialized, otherwise, there are random values in `fval`, `ival` etc. And `printExpression` will print messy.


mrzzzrm:
already included in #71

PR #65, opened 2017-09-05
Do not allow deprecated features for flex

timzimmermann:
In C++1z the `register` keyword was removed. Older versions of flex still use it, and the Makefile disabled those warnings.
In this PR, we switch to compile with C++1z and an up-to-date version of flex is used to compile the parser.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ install:
   - sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test
   - sudo apt-get -qq update
   - sudo apt-get install -y g++-6 libstdc++-6-dev
-  - sudo apt-get install -y flex valgrind
+  - sudo apt-get install -y valgrind
   - sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-6 90
   - sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-6 90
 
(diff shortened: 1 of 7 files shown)


timzimmermann:
Regarding the CI error:
There is a quick fix, but I don't like it. The fix would be to leave the Makefile as is and hope for contributors to use up-to-date software. That typically does not work. If they do not use up-to-date software, flex might generate C++1z-incompatible code.
Which brings us to the actual fix. The code generated by flex is checked into the repository so that people using the library do not necessarily need to install flex. The CI, however, generates the flex files on its own and compiles the library with those.
If we compile the parser with C++11 (which the Makefile currently does) we have two options:
1. Disable the warnings for `register` being deprecated (as in the current `master`). That would mean we will not notice automatically when code incompatible to C++1z is being introduced.
2. Enable the warnings, and compile with the flex code checked into the repo. We would need two compilation runs, because we would want to compile the flex code in the CI once as well to make sure it's correct.
If we compile the parser with C++1z (second commit) we need to tell clang to use the correct standard library (seems to be the reason for clang failure in the second commit). Specifying that in the Makefile leads to errors when compiling with gcc (locally), because gcc does not know that flag. However, due to the way [travis builds with two different compilers](https://docs.travis-ci.com/user/languages/cpp/), I do not know how we would specify that we only want to provide that flag for clang builds. We would probably also want to update the compiler versions for both gcc and clang because they currently use rather old ones.

How do we want to handle this going forward? Do we want to shift to C++1z, or do we want to go with the easy solution and ignore it in the CI, as we currently do?

mrks:
Disabling the warnings or having a broken CI don't sound like good options to me. What exactly is the flag that clang needs and gcc hates?

timzimmermann:
I agree. The flag is `-stdlib=libc++`.

I do not think I can really work on this after today, I am leaving tomorrow. So maybe you can delegate it to someone in the opossum fortnightly.

mrks:
Can you do something like

```
if [ "$CXX" = "clang++" ]; then export CXXFLAGS="-stdlib=libc++"; fi
```

As done here: https://stackoverflow.com/questions/22111549/travis-ci-with-clang-3-4-and-c11

Issue #69, opened 2017-09-25
Multiple "WHEN" conditions not supported in a "CASE" statement


agladshteyn:
The following statement has two "WHEN" conditions and an error ("syntax error, unexpected WHEN") is returned when trying to parse it by calling SQLParser::parseSQLString:

SELECT (CASE
	WHEN ("Calcs"."int0" = "NULL") OR ("Calcs"."int1" = "NULL") THEN "NULL"
	WHEN "Calcs"."int0" > "Calcs"."int1" THEN "Calcs"."int0"
	ELSE "Calcs"."int1" END) AS "TEMP_Test__1523549003__0_"
FROM "Calcs"

If the statement is changed to only contain one "WHEN" condition, then the parsing succeeds:
SELECT (CASE
	WHEN ("Calcs"."int0" = "NULL") OR ("Calcs"."int1" = "NULL") THEN "NULL"
	ELSE "Calcs"."int1" END) AS "TEMP_Test__1523549003__0_"
FROM "Calcs"

PR #75, opened 2018-01-15
Changing Grammar to extend CASE WHEN statement:

javrucebo:
Changing Grammar to extend CASE WHEN statement: 
this addresses also #69 

- allow multiple WHEN statements
- allow for syntax like `CASE x WHEN 1 THEN 2 WHEN 3 THEN 4 ELSE 5 END`
NOTE: This changes also the way the CASE operator is stored:
- CASE [expr] exprList [ELSE expr2] END
- exprList holds each of the WHEN statements with:
  expr := WHEN, expr2 := THEN

Added also tests in test/select_tests.cpp
and adapted the existing one to reflect the new storage


javrucebo:
Travis failing on the memory leak check. I do not know how to take care of this.

mrzzzrm:
Since I can't edit you branch, I have (hopefully) fixed your code on a different branch: https://github.com/hyrise/sql-parser/pull/76

Thank you for contributing!

PR #77, opened 2018-01-17
Add stringLength information to Statement

lawben:
This PR adds information about the length of the statement in the original string. This can be used, e.g. to split a multi statement query into the actual statement strings after parsing.

```cpp
auto sql = "SELECT * FROM foo; SELECT * FROM blub";

... parse here ...

assert(parse_result.getStatement(0)->stringLength == 18);
assert(parse_result.getStatement(1)->stringLength == 20);

auto foo_select = sql.substr(0, 18); 
auto blub_select = sql.substr(18, 20); 
``` 

This `stringLength` will contain the number of characters until the semicolon ending the statement or the end of the string. It will contain leading and trailing whitespace. The sum of all `stringLength`s should be `sql.length()`.

EDIT:
This also contains a few changes for OSX development. The default bison version on OSX is too old, so we allow the user to specify a custom bison path for compilation.


mrks:
Looks good to me, but @mrzzzrm knows more about this

Makefile
```diff
@@ -60,7 +60,7 @@ $(LIB_BUILD): $(LIB_OBJ)
 	$(LIBLINKER) $(LIB_LFLAGS) $(LIB_BUILD) $(LIB_OBJ)
 
 $(SRCPARSER)/flex_lexer.o: $(SRCPARSER)/flex_lexer.cpp $(SRCPARSER)/bison_parser.cpp
-	$(CXX) $(LIB_CFLAGS) -c -o $@ $< -Wno-sign-compare -Wno-unneeded-internal-declaration
+	$(CXX) $(LIB_CFLAGS) -c -o $@ $< -Wno-sign-compare -Wno-unneeded-internal-declaration -Wno-register
```
mrzzzrm: Added intentionally?

src/parser/parser_typedef.h
```diff
@@ -22,6 +22,9 @@ struct HSQL_CUST_LTYPE {
 
   int total_column;
 
+  // Length of the string in the SQL query string
+  int string_length;
```
mrzzzrm: any reason this is not size_t?

mrzzzrm:
Wouldn't claim I do, but the tests kinda prove you did something right.

src/parser/parser_typedef.h
lawben: no reason in particular. i just stuck with all the other variables being `int`

mrks: I think we can safely assume that 4 GB is sufficient for the SQL string ;)

Makefile
lawben: yes. bison uses `register` which has been deprecated and thus throws a compile on OSX warning resulting in an error. suggested fix is to use this flag.

src/parser/parser_typedef.h
mrzzzrm: You didn't just make that argument :D

Bensk1: :(

PR #79, opened 2018-01-18
Extend table alias with column renames: "AS tbl(c0, c1)"

f4lco:
Extend parser to handle aliases with column renames as found in SQL '92, named "correlation name" and "derived column list", respectively.


src/sql/Table.h
```diff
@@ -24,6 +24,14 @@ namespace hsql {
     char* name;
   };
 
+  struct Alias {
+    Alias(char *name, std::vector<char*>* columns = nullptr);
```
lawben: this should be `char* name`

src/sql/statements.cpp
```diff
@@ -231,6 +231,21 @@ namespace hsql {
     }
   }
 
+  // Alias
+  Alias::Alias(char *name, std::vector<char*>* columns) :
```
lawben: this should be `char* name`

src/sql/Table.h
f4lco: OK. Same for the member I guess.

src/sql/statements.cpp
f4lco: OK.

src/parser/bison_parser.y
```diff
@@ -730,7 +732,10 @@ literal_list:
 expr_alias:
 		expr opt_alias {
 			$$ = $1;
-			$$->alias = $2;
+			if ($2) {
+				$$->alias = strdup($2->name);
+				delete $2;
```
f4lco: On second thought the grammar is now screwed. Supporting both table and column aliases requires a deeper separation of concerns.
As we can see here, an alias for `expr` re-uses the general purpose alias rule, discarding possible columns.
As a result something like `select a + 1 as b(spam, eggs) from B;` would be successfully parsed in the current state. I'll attempt at a better solution.

f4lco:
I just got to work with TPCH-13 and fixed a minor alignment issue in the output routine of "sqlhelper.cpp".

mrks:
@mrzzzrm Could you please look at this?

f4lco:
@mrks why merge, let's rebase. I have rerun the code generation step, build's green now.

src/parser/bison_parser.y
f4lco: Clarification: done. Shouldn't have reused that particular grammar rule in the first place.

src/parser/bison_parser.y
```diff
@@ -941,16 +946,28 @@ table_name:
 	;
 
 
+table_alias:
+		alias
+	|	AS IDENTIFIER '(' ident_commalist ')' { $$ = new Alias($2, $4); }
+	;
+
+
+opt_table_alias:
```
mrzzzrm: still a big lack of understanding for bison on my side, but why do you have two separate rules opt_table_alias AND table_alias? Wouldn't opt_table_alias be sufficient?

f4lco: Technically one could merge `opt_x` and `x` rules, I do not know about a technical requirement. I would stick with separate clauses for the following reasons: 

1. The grammar rules are like building blocks. Possibly detailed control of mandatory and optional aliases is desired at a later stage. 
3. Inline with the rest of the grammar definitions.
4. It's clunky to nest optional clauses inside optional clauses. For instance, expressing `opt_table_alias` in one rule would result in the following:

        opt_table_alias:
    	    	opt_alias
    	    |	AS IDENTIFIER '(' ident_commalist ')' { $$ = new Alias($2, $4); }
    	    ;

    It does not seem to be as straightforward and obvious as the current proposal.

mrks: Yepp, that makes sense to me.

mrks:
I can't modify your PR. Could you either merge this with master or [give maintainers the right to edit your PRs?](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/)

```
[:~/Projekte/Opossum/Git/third_party/sql-parser] 2917502+ ± git push f4lco f4lco/table-alias-with-columns
Total 0 (delta 0), reused 0 (delta 0)
To github.com:f4lco/sql-parser.git
 ! [remote rejected] f4lco/table-alias-with-columns -> f4lco/table-alias-with-columns (permission denied)
error: failed to push some refs to 'git@github.com:f4lco/sql-parser.git'
```

f4lco:
Huh something is fishy. I always leave the "allow edits" checkbox set. Anyway I rebased the branch right now, merging should be feasible.

f4lco:
`git commit --allow-empty -m "trigger"` ;)

mrks:
Yepp, but as we have seen before, github somehow didn't want me to push to this branch.

PR #80, opened 2018-01-22
Allow '==' for operator equals

javrucebo:
This change allows '==' as operator for test of equality, like sqlite allows.
Although this is not according to the standard, it should be allowed here as sqlite is in widespread use.
See also: https://sqlite.org/lang_expr.html#collateop

### DIFF
--- a/src/parser/bison_parser.y
+++ b/src/parser/bison_parser.y
@@ -814,6 +814,7 @@ exists_expr:
 
 comp_expr:
 		operand '=' operand			{ $$ = Expr::makeOpBinary($1, kOpEquals, $3); }
+	|	operand EQUALS operand			{ $$ = Expr::makeOpBinary($1, kOpEquals, $3); }
 	|	operand NOTEQUALS operand	{ $$ = Expr::makeOpBinary($1, kOpNotEquals, $3); }
 	|	operand '<' operand			{ $$ = Expr::makeOpBinary($1, kOpLess, $3); }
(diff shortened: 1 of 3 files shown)


src/parser/bison_parser.y
```diff
@@ -814,6 +814,7 @@ exists_expr:
 
 comp_expr:
 		operand '=' operand			{ $$ = Expr::makeOpBinary($1, kOpEquals, $3); }
+	|	operand EQUALS operand			{ $$ = Expr::makeOpBinary($1, kOpEquals, $3); }
```
mrks: Is there a reason why we have two cases for `=` and `==` here?

javrucebo: There are not two cases. The point is that in `src/parser/flex_lexer.l` in current master there is no token definition for either `=` nor `==`. Therefore token `EQUALS` is empty.

Checking `operand '=' operand` is in the current `src/parser/bison_parser.y`. Through the pull request I am adding:
* In `src/parser/flex_lexer.l` the string `==` into the token `EQUALS`
* and adding that token above for bison.

An alternative solution could be to
* add `=` to token `EQUALS` in the flexer and
* change the bison expression from `operand '=' operand` to `operand EQUALS operand` (and not add my extra line).

mrks: I first thought that the second solution would be cleaner, but that would cause issues for UPDATE, where we cannot accept `SET x EQUALS y`.

PR #81, opened 2018-01-22
Adjust Join types

javrucebo:
- Eliminate LeftOuter and RightOuter and consolidate into Left/Right
- Rename kJoinOuter to kJoinFull as this is the better name
- Also adding "FULL JOIN", allowing still "OUTER JOIN" (not part of standard) and longer form "FULL OUTER JOIN"

This commit breaks the output format as kLeftOuter and kJoinOuter are deleted and kJoinOuter renamed to kJoinFull


mrks:
Yepp, this makes perfect sense. Thank you for the PR. Just out of curiosity - what brings you to look into this?

javrucebo:
I am trying to transition legacy sql code in a (semi)-automated way.

javrucebo:
@mrks please also update bison_parser.cpp/.h and flex_lexer.cpp/.h.
Although they are newly comiled through local `make` process, they should reflect the same right state on github.

PR #82, opened 2018-01-24
Remove unused enum members

mrzzzrm:
@mrks @javrucebo what do you think?


mrzzzrm:
@mrks I can't merge, so....

PR #83, opened 2018-01-26
Allow more variations of LIMIT/OFFSET

javrucebo:
In addtion to already supported LIMIT/OFFSET variants allow more to be parsed

Legacy:
  - LIMIT int
  - LIMIT int OFFSET int

Enhancement:
  - OFFSET int             (no limit)
  - LIMIT ALL              (no limit)
  - LIMIT NULL             (no limit)
  - LIMIT ALL OFFSET int   (no limit, but offset)
  - LIMIT NULL OFFSET int  (no limit, but offset)

Also ensures negative limits / offsets are set to kNoLimit/kNoOffset


src/parser/flex_lexer.h
```diff
@@ -92,13 +92,25 @@ typedef unsigned int flex_uint32_t;
 
 #endif /* ! FLEXINT_H */
 
-/* TODO: this is always defined, so inline it */
-#define yyconst const
+#ifdef __cplusplus
```
mrzzzrm: @javrucebo do you have any idea why these changes appeared? Different version of flex?

javrucebo: probably yes. I had previously not looked into any changes of the generated files by bison/flex and only commited the changes then when I have seen that the CI process tests both with the check-in version and with the re-generated one. obviously the first check then failed when not committing the changes to the generated bison/flex files

PR #84, opened 2018-01-26
Adding more complete support for set operations: UNION / INTERSECT / …

javrucebo:
Adding more complete support for set operations: UNION / INTERSECT / EXCEPT

Allow multiple operations, differentiate between the operators, honor parentheses

- internal data structure is changing. Instead of the `unionSelect` pointer
  to a second SelectStatement two new vectors as members of
  SelectStatement are introduced:

  - setStatement: holds pointers to SelectStatement(s) right of first
    set operation

  - setType: holds pointers to new struct SetDescription with members
    - type: Enum (kSetUnion, kSetIntersect, kSetExcept)
    - all: bool to indicate if keyword ALL was used

- if no parentheses are used then there is only one set of vectors, e.g. for
  `SELECT a1 FROM t1 UNION SELECT a2 FROM t2 EXCEPT ALL SELECT a3 FROM t3`
  then setStatement holds two SelectStatements
 ```
 SelectStatement
    Table t1
    ColumnRef a1
    setStatement
       SelectStatement
         Table t2
         ColumnRef a2
       SelectStatement
         Table t3
         ColumnRef a3
    setType
       SetDescription
         type kSetUnion
         all false
       SetDescription
         type kSetExcept
         all true
```

- if parentheses are used then nested structure is built e.g. for
  `SELECT a1 FROM t1 UNION (SELECT a2 FROM t2 EXCEPT ALL SELECT a3 FROM t3)`
```
  SelectStatement
    Table t1
    ColumnRef a1
    setStatement
       SelectStatement
         Table t2
         ColumnRef a2
         setStatement
           SelectStatement
             Table t3
             ColumnRef a3
         setType
           SetDescription
             type kSetExcept
             all true
    setType
      SetDescription
        type kSetUnion
        all false
```

### CODE REVIEW

File: src/sql/statements.cpp [hunk 1]
```diff
@@ -208,6 +215,34 @@ namespace hsql {
...
+
+    if (setType != nullptr) {
+      for (SetDescription* desc : *setType) {
+        delete desc;
+      }
+      delete setType;
+    }
+  }
+
+  SelectStatement* MakeOrAppendUnionList(SelectStatement* stmt,
```
mrzzzrm: It's not only a UnionList, is it? Maybe MakeOrAppendSetList?

test/select_tests.cpp
```diff
@@ -438,4 +438,76 @@ TEST(JoinTypes) {
...
   stmt = (SelectStatement*) result.getStatement(11);
   ASSERT_NULL(stmt->fromTable->join);
-}
\ No newline at end of file
+}
+
+TEST(SetOperations) {
+  SelectStatement* stmt;
+
+  TEST_PARSE_SQL_QUERY("select a from t1 union select b from t2; \
```
mrzzzrm: limit/order after set operations should probably included here? As in "SELECT a FROM t1 UNION (...) ORDER BY a LIMIT 5"

mrzzzrm:
The SELECT statement grammar could really use some doc or better names, but that's not your concern. The setType/setStatement architecture is a bit confusing (especially since "set" is such an ambiguous word) but I can see why you would use it. I guess if we increase the testing as requested this should be fine.

test/select_tests.cpp
javrucebo: I have added such tests for `WHERE`/`LIMIT`/`ORDER` as suggested

src/sql/statements.cpp
javrucebo: I have renamed to `MakeOrAppendSetList` as suggested

javrucebo:
I agree the naming could be confusing. Happy to change if you have other suggestion. As you pointed out I chose naming as those are operations on sets.

PR #85, opened 2018-01-31
Print matching token for operator types instead of plain integer

f4lco:
What is your opinion on the location of the stream operator on `OperatorType`? Currently it is an implementation detail of the utils. It would also possible to make the signature part of `Expr.h` to let other clients benefit from proper printing.
Should one be able to distinguish unary minus and binary minus?


src/util/sqlhelper.cpp
```diff
@@ -237,4 +228,44 @@ namespace hsql {
     }
   }
 
+  std::ostream& operator<<(std::ostream& os, const OperatorType& op) {
+    static const std::map<const OperatorType, const std::string> operatorToToken = {
```
mrzzzrm: maybe use a normal switch here? https://stackoverflow.com/questions/931890/what-is-more-efficient-a-switch-case-or-an-stdmap 

minor

f4lco: Interesting read. I would like to argue that this particular method is not subject to performance concerns. It's a debug helper for printing a parse tree directly to `cout`. Even larger query plans will contain a modest number of operators.
In this case I would favor conciseness and readability over lookup-performance. I think creating this map only once accounts for the performance aspect adequately.

f4lco:
OK I'll omit the commit with the generated files.

f4lco:
I really appreciated if you answered the questions from the PR comment, even if it's just explicit consent to the current state.

src/util/sqlhelper.cpp
mrks: As `operator<<` should never be a performance-critical method, I don't think that it makes sense to spend time here.

mrks:
Argh. No idea how this got lost. :(

f4lco:
@mrks @mrzzzrm finally! 😆

PR #94, opened 2018-05-22
No FROM required

### LINKED ISSUE #90: SELECT without FROM clause not supported
However, `SELECT 1+2;` is a perfectly valid SQL statement

mrzzzrm:
Fixes #90

This makes the FROM clause optional so you can use your DBMS as a calculator `SELECT 1+5-3`, e.g.


mrks:
Yay! We have the most overengineered calculator :D

PR #96, opened 2018-05-29
Add tests for SUBSTR support

timzimmermann:


mrzzzrm:
Curious, why does substr need dedicated tests? It's just another function, right?

timzimmermann:
It is, but it is now easier to deduce that the parser supports it.
For someone who is just curious about that and does not have an obsession with reading bison files, this can help.

timzimmermann:
@mrks CI had mercy, ready to merge.

PR #103, opened 2018-10-30
Added boolean literal support

alkim0:


mrks:
Mmmh, this is a difficult topic. SQL does not have a bool type and `TRUE`/`FALSE` are no literals (aka. `<value expression>`). Strictly speaking, the way I read the sql-1992 standard, it is illegal to write `... WHERE a = TRUE`. The only place where they are allowed is `... WHERE a > 100 IS TRUE`. Obviously, people sometimes use the form and virtually never the latter.

Furthermore, the standard requires a third bool type:

```
         <truth value> ::=
                TRUE
              | FALSE
              | UNKNOWN
```

We couldn't store that in the suggested bool.

The different databases do it their own way too:

- MySQL treats it as 0/1
- Oracle does not know it at all
- PostgreSQL has a non-standard BOOL column
- WebSQL and SQL.js do not know it

My suggestion would be to do it like MySQL and map true/false to the ints 1 and 0. This would allow people to use true/false in the way they probably intent to, should work for your use case, and should not lead to a lot of breaking changes.

alkim0:
How about this. Like you said, boolean literals are translated into ivals with expression type kExprLiteralInt, but the Expr struct is given an extra isBoolLiteral flag which identifies whether the expression came from a boolean literal or not. This:
1. By default treats booleans as int values like MySQL. 
2. Allows users who (like me) who want an explicit boolean data type in their database.
3. Allows future users who would like to implement UNKNOWN in the future to use an ival of 2 or -1 without the user getting confused as to whether the ival is supposed to be treated as an int or a boolean value.

I don't think not implementing TRUE/FALSE literals is an option given their widespread use. I think that would prevent sql-parser from parsing some real world workloads (such as mine).

mrks:
Yes, that sounds reasonable. As unknown would not be a literal, I would limit the `makeLiteral` modification to true and false. That way, we can simply use `makeLiteral(bool val);`.

Instead of updating this PR, I made some minimal changes in https://github.com/hyrise/sql-parser/commit/eb6c3ecd3c9493806402529e34570c84a9f43727 to that regard. If that sounds reasonable to you, please incorporate it in your PR and I'll merge it.

alkim0:
Done. Thanks!

PR #104, opened 2018-11-12
Support escaped strings (e.g., 'Max O''Mustermann')

mrks:


src/parser/flex_lexer.l
```diff
@@ -219,11 +224,11 @@ FALSE		TOKEN(FALSE)
...
 
-'[^'\n]*' {
-	// Crop the leading and trailing quote char
-	yylval->sval = hsql::substr(yytext, 1, strlen(yytext)-1);
-	return SQL_STRING;
-}
+\'                          { BEGIN singlequotedstring; strbuf = std::stringstream{}; }
+<singlequotedstring>\'\'    { strbuf << '\''; }
+<singlequotedstring>[^']*   { strbuf << yytext; }
+<singlequotedstring>\'      { BEGIN 0; yylval->sval = strdup(strbuf.str().c_str()); return SQL_STRING; }
```
mrzzzrm: just for me: what does BEGIN 0 do, roughly?

src/parser/flex_lexer.l
```diff
@@ -219,11 +224,11 @@ FALSE		TOKEN(FALSE)
...
+<singlequotedstring><<EOF>> { fprintf(stderr, "unterminated string\n"); return 0; }
```
mrzzzrm: Two lines below we prefix the error message with "[SQL-Lexer-Error] ". Feels weird to unconditionally report sql-parser errors to stderr, but we should be at least consistent with the error messages.

PR #110, opened 2019-03-19
Fix multi-threaded parsing

mrzzzrm:
This should fix multi-threaded parsing. Not a clean solution, I am guessing the following block should be a single, state-less pattern? @mrks

```
\'                          { BEGIN singlequotedstring; strbuf = std::stringstream{}; }
<singlequotedstring>\'\'    { strbuf << '\''; }
<singlequotedstring>[^']*   { strbuf << yytext; }
<singlequotedstring>\'      { BEGIN 0; yylval->sval = strdup(strbuf.str().c_str()); return SQL_STRING; }
<singlequotedstring><<EOF>> { fprintf(stderr, "[SQL-Lexer-Error] Unterminated string\n"); return 0; }
```

Anyway, the following hyrise playground does not crash anymore, so Toni should be able to work with this for the time being.

```c++
#include <iostream>
#include <thread>

#include "types.hpp"
#include "tpch/tpch_queries.hpp"
#include "SQLParser.h"

using namespace opossum;  // NOLINT

int main() {
  std::vector<std::thread> threads;

  for (size_t t = 0; t < 200; ++t) {
    threads.emplace_back([&]() {
      for (size_t p = 0; p < 10'000; ++p) {
        hsql::SQLParserResult result;
        hsql::SQLParser::parse(tpch_queries.at(19), &result);
        std::cout << "Parsing " << p << " is valid: " << result.isValid() << std::endl;
      }
    });
  }

  for (auto& thread : threads) {
    thread.join();
  }

  return 0;
}
```

### DIFF
--- a/src/parser/flex_lexer.cpp
+++ b/src/parser/flex_lexer.cpp
@@ -1,6 +1,6 @@
-#line 1 "flex_lexer.cpp"
+#line 2 "flex_lexer.cpp"
 
-#line 3 "flex_lexer.cpp"
+#line 4 "flex_lexer.cpp"
 
 #define  YY_INT_ALIGNED short int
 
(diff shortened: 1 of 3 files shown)


mrks:
Meh, if it works, I can't see anything wrong with it.

mrks:
Can't trigger Travis right now, but it looks like just a network issue.

toniSta:
Also works in my scenario. Good job :+1:

mrzzzrm:
@mrks travis gave its blessing, merge if you will

PR #111, opened 2019-03-31
Allow Expressions in LIMIT and OFFSET.

### LINKED ISSUE #91: LIMIT with Expression not supported
However `SELECT * FROM t LIMIT 1 + 2` works in Postgres

lanice:
Closes #91.

I modified the parser to allow hsql::Expr instead of just int values for LIMIT and OFFSET. Since this is my first time working with Bison, please take a close look in case I messed something up.
I also added some more tests to cover this change.

With this, the changes in Hyrise should be minimal to support this, right?

Oh and my Bison version (from the apt repository) apparently is lower than the previously used one (3.0.4 vs 3.1), is that a problem?


src/parser/bison_parser.y
```diff
@@ -724,18 +724,16 @@ opt_order_type:
...
 opt_limit:
-		LIMIT int_literal { $$ = new LimitDescription($2->ival, kNoOffset); delete $2; }
-	|	LIMIT int_literal OFFSET int_literal { $$ = new LimitDescription($2->ival, $4->ival); delete $2; delete $4; }
-	|	OFFSET int_literal { $$ = new LimitDescription(kNoLimit, $2->ival); delete $2; }
-	|	LIMIT ALL { $$ = nullptr; }
-	|	LIMIT NULL { $$ = nullptr;  }
-	|	LIMIT ALL OFFSET int_literal { $$ = new LimitDescription(kNoLimit, $4->ival); delete $4; }
-	|	LIMIT NULL OFFSET int_literal { $$ = new LimitDescription(kNoLimit, $4->ival); delete $4; }
+		LIMIT expr { $$ = new LimitDescription($2, Expr::makeNullLiteral()); }
+	|	LIMIT expr OFFSET expr { $$ = new LimitDescription($2, $4); }
```
mrzzzrm: why use NULL instead of nullptr? I do not know whether `LIMIT 5 OFFSET NULL` has a defined semantic, but should it really be indifferentiable from `LIMIT 5`? For MySQL, at least, `LIMIT 5 OFFSET NULL` is no valid input, but `LIMIT 5` is, so in order to be able to tell those two apart I think

`LIMIT 5 OFFSET NULL` -> $$ = new LimitDescription($2, Expr::makeNullLiteral()); 
`LIMIT 5` -> $$ = new LimitDescription($2, nullptr); 
`LIMIT ALL OFFSET expr` -> $$ = new LimitDescription(nullptr, $4);

lanice: Yup makes sense. I originally used nullptr actually, but changed it to Expr::Null, to have consistency with things like `LIMIT 5 OFFSET NULL`. But you're right, we actually want those to be different. I updated it now.

mrzzzrm:
@mrks u can merge now

mrks:
Can't you?

mrzzzrm:
Nope, not authorized

mrks:
Ah, thanks for letting me know. There were some unnecessary branch protections enabled.

PR #112, opened 2019-04-04
Fix print

mrks:
fixes #111


mrks:
@divag711 - Kindly confirm that this solves your issue.

divag711:
@mrks Yes, the print issue I was having works with the fix.

Issue #35, opened 2017-04-15
Use Makefile Generation Script (premake/autoconf)


torpedro:

mtangoo:
Here is my initial implementation of CMake. It is still rudimentary and supports only building library, not test nor example for now. Since it might be too long before I can make actual PR, I though I should leave the sources in case anyone wants to tke it further before then. Note sources are reorganized for better structure

mrks:
Nice, thank you for sharing this!

PR #126, opened 2019-05-24
Add all 99 TPC-DS queries to the test suite

julianmenzler:
This PR

- implements a `testSQLFiles(..)` in  `tpc_h_tests.cpp`
- renames `tpc_h_tests.cpp` into `tpc_benchmark_tests.cpp`
- adds 99 TPC-DS queries to the test suite – 53 of them can be parsed at the moment


test/queries/tpc-ds-01.sql
```diff
@@ -0,0 +1,22 @@
+with customer_total_return as
```
mrks: see the corresponding Hyrise PR. Any chance we can format this?

test/tpc_benchmark_tests.cpp
```diff
@@ -0,0 +1,239 @@
...
+
+  // Test ORDER BY clause.
+  ASSERT_NOTNULL(select20->order);
+  ASSERT_EQ(select20->order->size(), 1);
+  ASSERT(select20->order->at(0)->expr->isType(kExprColumnRef));
+  ASSERT_STREQ(select20->order->at(0)->expr->getName(), "S_NAME");
+}
+
+TEST(TPCDSGrammarTests) {
+  std::vector <std::string> files = {
```
mrks: Do we need this as a list or can we get this from the file system? Same for TPC-H.

test/tpc_benchmark_tests.cpp
```diff
@@ -0,0 +1,239 @@
...
+
+int testSQLFiles(std::string sqlDatasetName, std::vector <std::string> files) {
+
+  int testsFailed = 0;
+  std::string concatenated = "";
+  for (const std::string &file_path : files) {
+    std::string query = readFileContents(file_path);
+
+    concatenated += query;
+    if (concatenated.back() != ';') concatenated += ';';
```
mrks: Doc what you are doing here?

test/tpc_benchmark_tests.cpp
```diff
@@ -0,0 +1,239 @@
...
+      ++testsFailed;
+    } else {
+      mt::printOk(file_path.c_str());
+    }
+  }
+
+  SQLParserResult result;
+  SQLParser::parse(concatenated.c_str(), &result);
+  if (!result.isValid()) {
+    mt::printFailed(std::string(sqlDatasetName + "AllConcatenated").c_str());
```
mrks: Does the concatenated test really test something that would be untested before?

test/tpc_benchmark_tests.cpp
julianmenzler: No clue! But as far as I can tell, it works! 😃 
I just wrapped the old code within `testSQLFiles()`. So I'm not responsible for any of the magic here..

mrks: As far as I understand it, it also takes all sql files, builds a single string from it, and tries to parse that. Verifying that multiple statements in a single string work should be done in a separate test. I'd remove it.

test/queries/tpc-ds-01.sql
julianmenzler: Ok, have not processed any of those files. Just copied them over from @mweisgut. Do you know of a CLI tool for this job? These web formatters don't allow for batch processing.

mrzzzrm:
I like the idea of this - maybe it would be even better to merge this test with the tests that run `queries-good.sql`? It would be easier to add files to this in the future if there was a general-purpose "take all files in this directory and try to parse them"-test

test/tpc_benchmark_tests.cpp
julianmenzler: Ok. I'll look into that next week.

julianmenzler:
> I like the idea of this - maybe it would be even better to merge this test with the tests that run `queries-good.sql`? It would be easier to add files to this in the future if there was a general-purpose "take all files in this directory and try to parse them"-test

Yeah, @mrks thinks the same from what I read. The current approach an advantage though. We can specify the number of tests, which are allowed to fail. Either for TPC-H or TPC-DS:

```cpp
  std::vector <std::string> files = {
          "test/queries/tpc-ds-01.sql",
          "test/queries/tpc-ds-02.sql",
          "test/queries/tpc-ds-03.sql",
          ...
          "test/queries/tpc-ds-99.sql"
  };

  int testsFailed = testSQLFiles("TPCDS", files);
  ASSERT_EQ(testsFailed, 46);
```

mrzzzrm:
I don't see that as a decisive advantage. Queries that are "bad" get the `!SELECT...` prefix, that should be sufficient.

julianmenzler:
This would require some refactoring. But possible.

Btw.: I'm a little irritated by the memory leak failures which occurred in Travis-CI. On my local system, each check passes without any failures.

mrzzzrm:
Normally I would say: Use the current approach and open an issue for the possible improvements. But since we're not actively developing the parser, such an issue would probably remain untackled. Since nothing depends on this PR, I'd say the few hours to do this properly should be invested.

test/queries/tpc-ds-01.sql
mweisgut: I found [sqlformat](http://manpages.ubuntu.com/manpages/trusty/man1/sqlformat-3.1.html) but I haven't formatted the queries yet.

mweisgut: https://github.com/hyrise/hyrise/pull/1679/commits/a796a936af09d9ba40fa377219cd81ea3e325aba

julianmenzler: Nice, thanks a lot @mweisgut 👍

julianmenzler:
> I like the idea of this - maybe it would be even better to merge this test with the tests that run `queries-good.sql`? It would be easier to add files to this in the future if there was a general-purpose "take all files in this directory and try to parse them"-test

I implemented your suggested changes using the `filesystem` library.  Also, I had success testing everything locally. 
However, I encountered serious issues with header file and library paths regarding the `filesystem` library.

* Travis CI misses correct header files
* macOS requires me to explicitly set library paths (to the clang dir, installed by homebrew). These local paths shouldn't be part of the Makefile
* On sidon, everything compiles, but I ran into some segmentation faults. Probably a mismatch between header files and the linked library. @Bensk1 and you also experienced this problem in the past, he said.

My impression is that we won't find a clean Makefile solution to this problem. We should think about a transition to CMake for this project like @torpedro suggested in #35. In Hyrise, CMake successfully takes care of the described `filesystem` library problems.

mrks:
The compilers in Travis are ancient. Should we discuss our options in person?

julianmenzler:
> The compilers in Travis are ancient. Should we discuss our options in person?

Good idea. I'll come to your office. Probably Tuesday.

julianmenzler:
Today, @mrks and I had our discussion: We decided to defer this PR until we have CMake support in place. Since this PR does not include any parser functionality it is not a top priority.

PR #127, opened 2019-06-06
Support Create Table As Syntax

d-justen:


d-justen:
Mhh, but is `select_test.cpp` actually the right place to test this `CREATE TABLE` statement?

mrzzzrm:
sql_tests.cpp then

PR #133, opened 2019-11-14
Sort the input file list

bmwiedemann:
Sort the input file list
so that `libsqlparser.so.1` builds in a reproducible way
in spite of indeterministic filesystem readdir order.

See https://reproducible-builds.org/ for why this is good.

I tested that the other `find` result orders do not matter. Only .o file order in `LIB_OBJ` matters for the result.

This PR was done while working on reproducible builds for openSUSE.

also filed at https://github.com/envoyproxy/sql-parser/pull/1

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,7 +49,7 @@ else
 	LIB_CFLAGS  +=  -fPIC
 	LIB_LFLAGS = -shared -o
 endif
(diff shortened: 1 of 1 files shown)


mrks:
Haven't heard of this before; this is certainly a good initiative. Thanks!

PR #135, opened 2019-12-09
Fix build badge

mrks:
The badge referred to the outdated release (which has been deleted for a while now). This PR updates it so that the master is used.


mrks:
nevermind

PR #138, opened 2019-12-18
Add support to identify different set operators & allow chain of multiple set operators

Dencrash:


src/parser/bison_parser.y
```diff
@@ -607,13 +609,11 @@ select_statement:
 			$$ = $2;
 			$$->withDescriptions = $1;
 		}
-	|	opt_with_clause select_with_paren set_operator select_paren_or_clause opt_order opt_limit {
-			// TODO: allow multiple unions (through linked list)
-			// TODO: capture type of set_operator
-			// TODO: might overwrite order and limit of first select here
+	|	opt_with_clause select_with_paren set_operator select_statement_in_set_operation opt_order opt_limit {
```
Till-Later: What about a Query like this:
```
( SELECT * FROM A EXCEPT SELECT * FROM B ) EXCEPT SELECT * FROM C;
```
Wouldn't the "EXCEPT SELECT * FROM B " part simply get overwritten by the "EXCEPT SELECT * FROM C" part?

src/parser/bison_parser.y
```diff
@@ -752,12 +777,16 @@ opt_top:
 	;
 
 opt_limit:
-		LIMIT expr { $$ = new LimitDescription($2, nullptr); }
+	limit { $$ = $1; }
+	|	/* empty */ { $$ = nullptr; }
+	;
+
+limit:
```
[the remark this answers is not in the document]
Dencrash: Well yeah, that was a left over from one of our first tries that I did not remove for this pr. TBH I almost feel like it would help the understanding for people to have a difference here. Will talk about it with our TA on wednesday

src/parser/bison_parser.y
```diff
@@ -725,10 +747,13 @@ opt_having:
 	|	/* empty */ { $$ = nullptr; }
 
 opt_order:
-		ORDER BY order_list { $$ = $3; }
+		order { $$ = $1; }
 	|	/* empty */ { $$ = nullptr; }
 	;
 
+order:
```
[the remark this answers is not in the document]
Dencrash: Well yeah, that was a left over from one of our first tries that I did not remove for this pr. TBH I almost feel like it would help the understanding for people to have a difference here. Will talk about it with our TA on wednesday

src/parser/bison_parser.y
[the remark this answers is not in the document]
Dencrash: True. I did not look into this case. The only solution I came up with over the break will  complicate the structure of unrelated queries though. Will have to talk about this at our meeting

src/parser/bison_parser.y
```diff
@@ -662,18 +668,34 @@ select_no_paren:
...
+		$$->isAll = $2;
+		}
 	;
 
 set_type:
-		UNION
-	|	INTERSECT
-	|	EXCEPT
+		UNION {
+		$$ = new SetOperator();
```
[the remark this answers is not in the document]
Dencrash: True in general. I believe I did that for consistency sake, since the rest of the codebase works the same way

src/parser/bison_parser.y
Dencrash: True, same problem as above though. Solution will complicate some unrelated queries. Will see if this desirable

Dencrash:
Thanks for the comments and spotting some of the oversights!

src/parser/bison_parser.y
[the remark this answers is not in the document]
Dencrash: Fixed in new patch

src/parser/bison_parser.y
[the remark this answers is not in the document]
Dencrash: Fixed in new patch

src/parser/bison_parser.y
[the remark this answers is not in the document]
Dencrash: Fixed in new patch

mrks:
What's the status on this? Once the current master is merged, is this ready for a final review?

Dencrash:
@mrks that depends. There is one bug still in there that is not yet closed. The bug is not something we added but already exists for the current system. The solution for that is relatively easy, but ugly as sin and we were stalling last week to see, if we can find a better solution.

We could do the following:
Let every query have a vector of set operation instead of only allowing one linked and nested set operation. This would mean that we don't overwrite old set operations anymore when specific parenthesis are used, but it means that we mix a vector with linked lists.

This is the only thing that is currently not done.

mrks:
Can't you iterate over the chained list of select statements and add the set operation to the end? Something like

```
$$ = $1;
while ($$->setOperator) $$ = $$->setOperator;
$$->setOperator = $2;
```

No idea if bison likes that though.

Also, could you add some tests with similar (multiple, different, nested, ...) set operations in the WITH part of the query?

Dencrash:
Hanging onto the event would mean that we have the wrong order of operation. We also cannot use the other Select statement because the set operation there could also be overwritten.

We can add test with WITH Statements and more nested queries

mrks:
Let's discuss this on Wednesday.

src/parser/bison_parser.y
Dencrash: Fixed in newest Patch

Dencrash:
@mrks I have added an explanation to the evaluation order of SetOperators as a comment. From our side this should be ready for your final review

src/parser/bison_parser.y
```diff
@@ -127,6 +127,7 @@ int yyerror(YYLTYPE* llocp, SQLParserResult* result, yyscan_t scanner, const cha
 	hsql::GroupByDescription* group_t;
 	hsql::UpdateClause* update_t;
 	hsql::Alias* alias_t;
+	hsql::SetOperator* set_operator_t;
```
mrks: example_bin should not be committed

src/parser/bison_parser.y
```diff
@@ -689,33 +691,40 @@ select_statement:
...
 			$$->withDescriptions = $1;
 		}
-	|	opt_with_clause select_with_paren set_operator select_paren_or_clause opt_order opt_limit {
-			// TODO: allow multiple unions (through linked list)
-			// TODO: capture type of set_operator
-			// TODO: might overwrite order and limit of first select here
+	|	opt_with_clause select_with_paren set_operator select_statement_in_set_operation opt_order opt_limit {
 			$$ = $2;
+			if ($$->setOperators == nullptr) {
+				$$->setOperators = new std::vector<SetOperator*>();
```
mrks: Is this really consistent with what we do in other places? I feel that pointers are overused in the parser anyway, but if we set it to an empty vector here and never have to check if this is a nullptr (i.e., unset), what is the point of the indirection?

src/parser/flex_lexer.l
```diff
@@ -131,6 +131,8 @@ OUTER		TOKEN(OUTER)
 RIGHT		TOKEN(RIGHT)
 TABLE		TOKEN(TABLE)
 UNION		TOKEN(UNION)
+INTERSECT	TOKEN(INTERSECT)
+EXCEPT		TOKEN(EXCEPT)
```
mrks: Whitespace

src/util/sqlhelper.cpp
```diff
@@ -168,9 +168,36 @@ namespace hsql {
       }
     }
 
-    if (stmt->unionSelect != nullptr) {
-      inprint("Union:", numIndent + 1);
-      printSelectStatementInfo(stmt->unionSelect, numIndent + 2);
+    if (stmt->setOperators != nullptr) {
+      for (SetOperator* setOperator : *stmt->setOperators) {
+        if (setOperator->setType == SetType::kSetUnion) {
```
mrks: Why not a switch?

test/sql_tests.cpp
```diff
@@ -301,6 +298,174 @@ TEST(StringLengthTest) {
   ASSERT_EQ(result.getStatement(2)->stringLength, 21);
 }
 
+TEST(ExceptOperatorTest) {
```
mrks: I think a test where the last SELECT has its own ORDER/LIMIT in addition to the global one would help avoid future problems.

src/sql/SelectStatement.h
```diff
@@ -58,7 +75,19 @@ namespace hsql {
...
+    // have SetOperator connections to other SelectStatements.
+    // To evaluate the operators in the correct order: 
+    //    1. Fully evaluate the nested SelectStatement
+    //    2. Connect the nested SelectStatement to our original
+    //       with the given SetOperation
+    //    3. Connect the result with the next SetOperator in the
+    //       vector
+    //    4. Repeat until all SetOperators in the vector are
+    //       connected 
+    std::vector<SetOperator*>* setOperators;
```
mrks: How do you feel about „setDefinition(s)“?

src/sql/SelectStatement.h
```diff
@@ -58,7 +75,19 @@ namespace hsql {
...
+    // different SelectStatement. This statement can itself
+    // have SetOperator connections to other SelectStatements.
+    // To evaluate the operators in the correct order: 
+    //    1. Fully evaluate the nested SelectStatement
+    //    2. Connect the nested SelectStatement to our original
+    //       with the given SetOperation
+    //    3. Connect the result with the next SetOperator in the
+    //       vector
+    //    4. Repeat until all SetOperators in the vector are
+    //       connected
```
mrks: I‘m not sure if someone who didn‘t work through this knows what you mean with „connect“. Also, what is „our original“? Maybe give an example and tell people how to understand it? I‘d also include the outer order/limit.

src/parser/bison_parser.y
```diff
@@ -809,7 +830,7 @@ opt_having:
 	;
 
 opt_order:
-		ORDER BY order_list { $$ = $3; }
+		ORDER BY order_list { $$ = $3; };
```
mrks: Why? The diff above doesn‘t seem to have this either.

src/parser/bison_parser.y
```diff
@@ -689,33 +691,40 @@ select_statement:
...
-			// Limit could have been set by TOP.
-			if ($6 != nullptr) {
-				delete $$->limit;
-				$$->limit = $6;
-			}
+select_statement_in_set_operation:
+		select_with_paren
+	|	select_part_of_set_operation;
+
+select_part_of_set_operation:
```
mrks: Terminology: What is the different between a select_statement_in_set_operation and a select_part_of_set_operation?

src/util/sqlhelper.cpp
Dencrash: Sure

src/parser/flex_lexer.l
Dencrash: The file is tab formated...I'm not quite sure how to best aligne it. Currently it's aligned for tab_width >= 4. 

Or do you mean something else for Whitespace?

test/sql_tests.cpp
Dencrash: Sure

src/parser/bison_parser.y
Dencrash: It's consistent in the sense that all the other vectors are also initialized in the parser_definition and not during initialization (afaik). The nullptr check is not necessary for any other rule

src/sql/SelectStatement.h
Dencrash: Only for the variable name or also the class names?

src/sql/SelectStatement.h
Dencrash: Okay

src/parser/bison_parser.y
Dencrash: True

src/parser/flex_lexer.l
mrks: Weird. Now it looks correct. No idea what github showed me. nevermind

src/parser/bison_parser.y
Dencrash: A select_statement_in_set_operation can either have a select statement in parenthesis or one that is not in parenthesis. 

select_part_of_set_operation is the case in which we don't have parenthesis and then we need to make sure that there is no order or limit term compared to a normal statement

src/parser/bison_parser.y
mrks: Are we talking about the same thing? I see a couple of places where a vector is explicitly nullptr:

```
opt_column_list:
		'(' ident_commalist ')' { $$ = $2; }
	|	/* empty */ { $$ = nullptr; }
	;
```

```
opt_order:
		ORDER BY order_list { $$ = $3; }
	|	/* empty */ { $$ = nullptr; }
	;

order_list:
		order_desc { $$ = new std::vector<OrderDescription*>(); $$->push_back($1); }
	|	order_list ',' order_desc { $1->push_back($3); $$ = $1; }
	;
```

It looks like `vector<[^(]*>\(\)` is usually followed by a push_back.

src/sql/SelectStatement.h
mrks: I don't see a reason for the two not to be equal.

src/parser/bison_parser.y
mrks: Mmh, maybe `select_within_set_(operation|definition)` and `select_within_set_definition_no_parentheses`? Also, those two explanations would help as a comment.

Dencrash: Seems good to me

src/sql/SelectStatement.h
Dencrash: Then I'd maybe prefer "setOperatorDefinitions" it's getting a bit verbose but seems to get closer to what we are trying to say

src/parser/bison_parser.y
Dencrash: True, semicolon seems to have sneaked in there

src/sql/SelectStatement.h
mrks: I guess I don't like the "operator" part of it. `setOperation` would be slightly better, too.

*There are 2 hard problems in computer science: cache invalidation, naming things, and off-by-1 errors.*

src/parser/bison_parser.y
Dencrash: Well the order list is initially set as nullptr as well. They just don't need the check because they have a very specific base case.

I'm fully in favor of doing the initialization in the constructor, however it would be inconsistent with the rest of the code-base. If you give the okay for the inconsistency, I have no problem to change it

src/sql/SelectStatement.h
Dencrash: Well, this is not helped by the fact that setAnything basically sounds like a setter for stuff. That's why I like the operator behind it to make it clearer

Dencrash: I'd say `setOperation` is best then

src/parser/bison_parser.y
mrks: Are you in the office right now? Maybe this is easier to discuss offline.

Dencrash: Saw this too late. You seem to be gone now...

Are you working this weekend again?

mrks: Yes, but not in the office

Dencrash:
I think that I added all the requested changes. Let me know, if you think this is ready to merge

Daniel-Xu:
Hi, @Dencrash. From the test case, I don't see that the `setOperations` has multiple children. Is there any case that requires the `setOperations` to be a vector instead of a `selectOption` pointer?

Daniel-Xu:
From the comment in the source code. 
I believe `(SELECT * FROM students INTERSECT SELECT * FROM students_2) UNION SELECT * FROM students_3 ORDER BY grade ASC;` this is the case I was talking about ↑

mweisgut:
> Hi, @Dencrash. From the test case, I don't see that the `setOperations` has multiple children. Is there any case that requires the `setOperations` to be a vector instead of a `selectOption` pointer?

Do you mean a `setOperation` pointer rather than a `selectOption` pointer?

I think you're right. Thank's for pointing to that issue. #199 contains a patch addressing it.

Daniel-Xu:
> > Hi, @Dencrash. From the test case, I don't see that the `setOperations` has multiple children. Is there any case that requires the `setOperations` to be a vector instead of a `selectOption` pointer?
> 
> Do you mean a `setOperation` pointer rather than a `selectOption` pointer?
> 
> I think you're right. Thank's for pointing to that issue. #199 contains a patch addressing it.

yea, selectOption was a typo

klauck:
Hi @Daniel-Xu , hi @mweisgut,
if ```setOperations``` is only a pointer, we cannot express the order of set operations anymore,
e.g., there is no way to express the difference between ```(Q1 EXCEPT Q2) EXCEPT Q3``` and ```Q1 EXCEPT (Q2 EXCEPT Q3)```.
Using this example: For ```(Q1 EXCEPT Q2) EXCEPT Q3```, Q1 must have two children and ```setOperations``` must be a vector. For ```Q1 EXCEPT (Q2 EXCEPT Q3)```, Q1 and Q2 have one child each.

klauck:
Nevertheless, in the current version ```Q1 EXCEPT Q2 EXCEPT Q3``` == ```Q1 EXCEPT (Q2 EXCEPT Q3)```, which I think is a bug.
It should be ```Q1 EXCEPT Q2 EXCEPT Q3``` == ```(Q1 EXCEPT Q2) EXCEPT Q3```.

klauck:
@mweisgut pointed out that INTERSECT binds more tightly than UNION and EXCEPT. [https://www.postgresql.org/docs/14/queries-union.html]
This is not yet implemented in the parser.

Daniel-Xu:
> Hi @Daniel-Xu , hi @mweisgut, if `setOperations` is only a pointer, we cannot express the order of set operations anymore, e.g., there is no way to express the difference between `(Q1 EXCEPT Q2) EXCEPT Q3` and `Q1 EXCEPT (Q2 EXCEPT Q3)`. Using this example: For `(Q1 EXCEPT Q2) EXCEPT Q3`, Q1 must have two children and `setOperations` must be a vector. For `Q1 EXCEPT (Q2 EXCEPT Q3)`, Q1 and Q2 have one child each.

Thanks for the response, this makes sense now.

PR #139, opened 2020-01-06
Add COPY for import and export 

dey4ss:
refs [hyrise/hyrise#1933](https://github.com/hyrise/hyrise/issues/1933)


src/parser/bison_parser.y
```diff
@@ -215,6 +218,7 @@ int yyerror(YYLTYPE* llocp, SQLParserResult* result, yyscan_t scanner, const cha
 %type <group_t>		    opt_group
 %type <alias_t>		    opt_table_alias table_alias opt_alias alias
 %type <with_description_t>  with_description
+%type <import_type_t>	opt_file_type import_file_type
```
mrks: Could you harmonize the mix of `file_type`, `import_file_type`, and `import_type`?

src/parser/bison_parser.y
```diff
@@ -374,24 +383,65 @@ execute_statement:
 
 /******************************
  * Import Statement
+ * IMPORT FROM TBL FILE 'test/students.tbl' INTO students
+ * COPY students FROM 'test/students.tbl' (WITH FORMAT TBL)
```
mrks: Did we already use parentheses for optional parts in this file? Otherwise, brackets [] would be more common.

src/parser/bison_parser.y
```diff
@@ -374,24 +383,65 @@ execute_statement:
...
 	;
 
 import_file_type:
-		CSV { $$ = kImportCSV; }
+		IDENTIFIER {
+			if (strcasecmp($1, "csv") == 0) {
+			 	$$ = kImportCSV;
+			} else if (strcasecmp($1, "tbl") == 0) {
+			 	$$ = kImportTbl;
+			} else if (strcasecmp($1, "binary") == 0) {
```
mrks: Is 'binary' a postgres thing? Do we want to support 'bin' as well, given that that is the file extension?

src/parser/bison_parser.y
```diff
@@ -374,24 +383,65 @@ execute_statement:
...
+	|	/* empty */  { $$ = kImportAuto; }
+	;
+
+
+/******************************
+ * Export Statement
+ * COPY students TO 'test/students.tbl' (WITH FORMAT TBL)
+ ******************************/
+export_statement:
+		COPY table_name TO file_path opt_file_type {
```
mrks: As this uses opt_file_type, too, it is weird that opt_file_type has a import_file_type in an export context

src/parser/sql_keywords.txt
```diff
@@ -76,10 +76,11 @@ VALUES
 DIRECT
 SORTED
 
+COPY
```
mrks: Does this file serve a purpose?

src/sql/ExportStatement.h
```diff
@@ -0,0 +1,21 @@
+#ifndef SQLPARSER_EXPORT_STATEMENT_H
+#define SQLPARSER_EXPORT_STATEMENT_H
+
+#include "ImportStatement.h"
+#include "SQLStatement.h"
+
+namespace hsql {
+  // Represents SQL Export statements.
+  struct ExportStatement : SQLStatement {
+    ExportStatement(ImportType type);
```
mrks: ImportType?

src/parser/sql_keywords.txt
dey4ss: I don't know, I just thought it is somehow common to update it

src/parser/bison_parser.y
dey4ss: Exactly, Postgres knows text, csv and binary as formats. I think bin is also a quite handy option, so I'm gonna add it

src/parser/bison_parser.y
dey4ss: I myself would prefer using FileType instead of ImportType in general for both Import and Export statement (so it would sense for both). I just thought renaming it could be a thing regarding downwards compatibility.

mrks: Ok, in that case, please add a comment.

dey4ss: Do I get it right and renaming `ImportType` to `FileType` and altering its values to `kFileCSV`, ... with just adding a comment there that this has been done is okay?

mrks: My thought was to leave it as it, but add comments where appropriate that this is done for compat reasons

dey4ss: Alright, are those two comments enough?

PR #145, opened 2020-05-24
Use awk to check versions

### LINKED ISSUE #144: Making of the parser fails on Centos because of bc
Some Linux distributions (e.g. Centos) don't have bc installed by default. And some (e.g Centos 8) don't have binary packages for it available either. It looks like it's not a very good idea to use bc to check versions of bison and flex in the parser Makefile.

msiomkin:
Some Linux distributions (e.g. Centos) don't have bc installed by
default. And some (e.g Centos 8) don't have binary packages for it
available either. It's safer to use awk instead of bc to check
versions of bison and flex in the parser Makefile.

Closes #144


mrks:
Great, thank you!

PR #150, opened 2020-06-03
Refine example in README.md

TennyZhuang:
Signed-off-by: TennyZhuang <zty0826@gmail.com>

fix a compile error in `if (statement->isType(hsql::kSelectStmt))` and refine cast syntax.


mrks:
Thank you!

PR #153, opened 2020-06-16
Add support of CAST(expression AS datatype(length))

### LINKED ISSUE #154: CAST(expression AS datatype(length)) support
CAST(expression AS datatype(length)) is a useful construction and is a part of ANSI SQL-92 standard. But the parser doesn't support it.

fixed by #153

msiomkin:
CAST(expression AS datatype(length)) construction is a part of
ANSI SQL-92 standard. So we have to support it.

fixes #154


mrks:
Thanks a lot!

PR #156, opened 2020-09-01
Fix segfault on printing cast/extract exprs

ankushrayabhari:
Previously, printing a cast/extract expression caused a segfault as the logic for `kExprFunctionRef` prints the `exprList` field which is not set for those types. Fixed this by adding customized printing logic for those expression types.


mrks:
Could you update this for #157, please?

I was contemplating whether adding new expression types is really the best way to go, given that it is a breaking change. However, I agree that it is weird that some use exprList and others use datetimeField/columnType.

mrks:
We wonder if semver makes sense if we are only supporting one version at a time. We'd love to turn this into a stable API, but we currently lack the resources for that. :/

PR #164, opened 2020-11-30
Fixing print ORDER BY bug: only first sort column was printed.

bettojanea:
Issue is in sqlhelper.cpp, method printSelectStatementInfo.


mrks:
Thanks for catching this. Your PR did not yet handle `SELECT * FROM test UNION SELECT * FROM test2 ORDER BY a DESC, b ASC;`, which I have added.

PR #167, opened 2021-02-15
Missing move assignment

cdmh:
`SQLParserResult` had a move ctor but no move assignment operator, which can lead to unexpected semantics


mrks (approved):
Interesting, haven't seen the pattern of reusing the assignment for the constructor yet. Neat.

Issue #61, opened 2017-08-16
Segment fault with example while using |create table statement|


vivekgalatage:
I built the code on Mac and tried executing the `example` with below statement

```
./example "CREATE TABLE Persons (PersonID int);"
```
And this results in a SEGFAULT. So I tried to debug and got the following stack

```
$ lldb example
(lldb) target create "example"
Current executable set to 'example' (x86_64).
(lldb) r "CREATE TABLE Persons (PersonID int);"
Process 45712 launched: '/Users/user/workspace/sql-parser/example/example' (x86_64)
Parsed successfully!
Number of statements: 1
CreateStatment
	Persons
Process 45712 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
    frame #0: 0x00007fffdae6ab52 libsystem_c.dylib`strlen + 18
libsystem_c.dylib`strlen:
->  0x7fffdae6ab52 <+18>: pcmpeqb (%rdi), %xmm0
    0x7fffdae6ab56 <+22>: pmovmskb %xmm0, %esi
    0x7fffdae6ab5a <+26>: andq   $0xf, %rcx
    0x7fffdae6ab5e <+30>: orq    $-0x1, %rax
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
  * frame #0: 0x00007fffdae6ab52 libsystem_c.dylib`strlen + 18
    frame #1: 0x0000000100092ccd libsqlparser.so`hsql::inprint(char const*, unsigned long) + 93
    frame #2: 0x0000000100000c19 example`main + 489
    frame #3: 0x00007fffdae34235 libdyld.dylib`start + 1
    frame #4: 0x00007fffdae34235 libdyld.dylib`start + 1
```

ilsken:
This turned out to be a problem in the `printCreateStatementInfo` function in `src/util/sqlhelper.cpp`. The value of `stmt->fileName` was null so it errored out.

This modified version adds a null check and a loop to print the column names and types. It's really ugly but I hastily put this together just to try and get this working:

```
  void printCreateStatementInfo(const CreateStatement* stmt, uintmax_t numIndent) {
    inprint("CreateStatment", numIndent);
    inprint(stmt->tableName, numIndent + 1);
    if (stmt->filePath) {
        inprint(stmt->filePath, numIndent + 1);
    }
[...]

Issue #113, opened 2019-04-04
Properly test the print methods


mrks:
As seen in #111, the print* methods are entirely untested. Since we want this to be useful for people outside of Hyrise, too, we should make sure that we add tests for this (and any other untested parts).

PR #179, opened 2021-07-16
Extend printCreateStatementInfo: add column definitions

klauck:
Add column definitions for printCreateStatementInfo as suggested as part of #61

### DIFF
--- a/src/util/sqlhelper.cpp
+++ b/src/util/sqlhelper.cpp
@@ -301,6 +301,12 @@ namespace hsql {
     inprint("CreateStatement", numIndent);
     inprint(stmt->tableName, numIndent + 1);
     if (stmt->filePath) inprint(stmt->filePath, numIndent + 1);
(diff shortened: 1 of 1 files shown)


mweisgut:
If we parse, for example, `CREATE TABLE t1 (c1 INTEGER, c2 INTEGER);`, the implementation of this PR results in the following print output:
```
CreateStatement
	t1
		c1
			INT
		c2
			INT
```

Something like the following would be more descriptive, IMHO:
```
CreateStatement
	Type:
		Table
	Name:
		t1
	Columns:
		c1
			INT
		c2
			INT
```

Since a `CreateStatement` can have different `CreateType`s, adding the type to the print output would be beneficial.

Furthermore, we should pay attention to the test coverage for new code and add tests correspondingly IMO.

klauck:
Yes, we should address #113 first.
Following, I will think about the printed information.

Issue #19, opened 2016-09-04
Distinct not handled in sqlhelper.cpp


Achilles-96:
In `sqlhelper.cpp`, distinct flag is not checked. So, when running example, distinct is not shown in the output.

PR #180, opened 2021-07-16
Klauck/print select distinct

klauck:
Check distinct flag for printSelectStatementInfo as suggested in #19

### DIFF
--- a/src/util/sqlhelper.cpp
+++ b/src/util/sqlhelper.cpp
@@ -182,6 +182,9 @@ namespace hsql {
 
   void printSelectStatementInfo(const SelectStatement* stmt, uintmax_t numIndent) {
     inprint("SelectStatement", numIndent);
+    if (stmt->selectDistinct) {
+      inprint("distinct", numIndent + 1);
+    }
     inprint("Fields:", numIndent + 1);
     for (Expr* expr : *stmt->selectList) printExpression(expr, numIndent + 2);
 
--- a/src/util/sqlhelper.cpp
+++ b/src/util/sqlhelper.cpp
@@ -301,6 +301,12 @@ namespace hsql {
     inprint("CreateStatement", numIndent);
     inprint(stmt->tableName, numIndent + 1);
     if (stmt->filePath) inprint(stmt->filePath, numIndent + 1);
+    if (stmt->columns) {
+      for (ColumnDefinition* column: *stmt->columns) {
+        inprint(column->name, numIndent + 2);
+        inprint(column->type, numIndent + 3);
+      }
+    }
   }
 
   void printInsertStatementInfo(const InsertStatement* stmt, uintmax_t numIndent) {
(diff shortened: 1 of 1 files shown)


src/util/sqlhelper.cpp
```diff
@@ -301,6 +304,12 @@ namespace hsql {
     inprint("CreateStatement", numIndent);
     inprint(stmt->tableName, numIndent + 1);
     if (stmt->filePath) inprint(stmt->filePath, numIndent + 1);
+    if (stmt->columns) {
```
mweisgut, line 307: Since improving the `printCreateStatementInfo` does not only address the `print select distinct` issue (#19), I would change the PR description respectively. Accordingly, the title of the PR should be descriptive and reflect the content of the PR.

In addition, if we parse, for example, `CREATE TABLE t1 (c1 INTEGER, c2 INTEGER);`, the implementation of this PR results in the following print output:
```
CreateStatement
	t1
		c1
			INT
		c2
			INT
```

Something like the following would be more descriptive, IMHO:
```
CreateStatement
	Type:
		Table
	Name:
		t1
	Columns:
		c1
			INT
		c2
			INT
```

Since a `CreateStatement` can have different `CreateType`s, adding the type to the print output would be beneficial.

mweisgut:
Even though the existing print functions are not tested yet, we should pay attention to the test coverage for new code and add tests correspondingly IMO.

src/util/sqlhelper.cpp
```diff
@@ -182,6 +182,9 @@ namespace hsql {
 
   void printSelectStatementInfo(const SelectStatement* stmt, uintmax_t numIndent) {
     inprint("SelectStatement", numIndent);
+    if (stmt->selectDistinct) {
+      inprint("distinct", numIndent + 1);
```
mweisgut, line 186: I'm not sure about the scope of this PR. If the scope is not only printing the `distinct` information but also improving the `printSelectStatementInfo`, printing `SelectStatement::withDescriptions` is also still missing in `printSelectStatementInfo`.

src/util/sqlhelper.cpp
mweisgut: I just saw #179, so I think the changes in `printCreateStatementInfo` are not intended to be in this PR?

src/util/sqlhelper.cpp
cdmh: there is also `hsql::Expr::distinct` for cases such as `SELECT COUNT(DISTINCT column) ...`. If `DISTINCT` is going to be output, shouldn't this case be too?

src/util/sqlhelper.cpp
klauck: Yes, I revert these changes

cdmh:
there is also `hsql::Expr::distinct` for cases such as `SELECT COUNT(DISTINCT column) ...`. If `DISTINCT` is going to be output, shouldn't this case be too?

PR #189, opened 2021-09-09
Report an error if the valgrind leak check cannot be executed

### LINKED ISSUE #188: Memory leak check reported as success when it was not executed
In `test.sh`, the memory leak check is executed with valgrind. However, we dump the output to `/dev/null`. Thus, even if valgrind is actually not installed and the memory check cannot be executed (`valgrind: command not found` is dumped to `/dev/null`), the resulting message is `Memory leak check succeeded!`.

We should report a warning / error message if the vagrind memory check cannot be executed.

mweisgut:
fixes #188

### DIFF
--- a/test/test.sh
+++ b/test/test.sh
@@ -46,14 +46,8 @@ elif [ $MEM_LEAK_RET -eq 200 ]; then
 elif [ $MEM_LEAK_RET -eq 127 ]; then
   printf "${RED}Memory leak check failed: command 'valgrind' not found!${NC}\n"
 else
-  printf "${RED}Memory leak check failes: error code ${MEM_LEAK_RET}!${NC}\n"
+  printf "${RED}Memory leak check failed: error code ${MEM_LEAK_RET}!${NC}\n"
 fi
-
-#else
-#  MEM_LEAK_RET=1
-#  RET=1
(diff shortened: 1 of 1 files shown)


test/test.sh
```diff
@@ -32,18 +32,29 @@ fi
...
+  RET=0
+elif [ $MEM_LEAK_RET -eq 200 ]; then
+  printf "${RED}Memory leak check failed!${NC}\n"
+elif [ $MEM_LEAK_RET -eq 127 ]; then
+  printf "${RED}Memory leak check failed: command 'valgrind' not found!${NC}\n"
 else
-	MEM_LEAK_RET=1
-	RET=1
-	printf "${RED}Memory leak check failed!${NC}\n"
+  printf "${RED}Memory leak check failes: error code ${MEM_LEAK_RET}!${NC}\n"
```
Bensk1, line 49: faile**d**

test/test.sh
```diff
@@ -32,18 +32,29 @@ fi
...
 fi
 
+#else
+#  MEM_LEAK_RET=1
+#  RET=1
+#  printf "${RED}Memory leak check could not be executed!${NC}\n"
+#fi
+
```
Bensk1, line 57: remove?

Bensk1 (changes requested):

Issue #184, opened 2021-08-09
DATE subtract with INTERVAL is not parsed


cdmh:
This query from  `tpc-h` doesn't parse, and reports an error at column 75, inside `interval '75'`.
```SQL
SELECT * FROM lineitem where l_shipdate <= date '1998-12-01' - interval '75' day (3);
```

This code
```C++
    REQUIRE(hsql::SQLParser::parse(
        "SELECT * FROM lineitem where l_shipdate <= date '1998-12-01' - interval '75' day (3);",
        &parser_result) == true);
    if (!parser_result.isValid())
        std::cerr << parser_result.errorMsg() << " at column " << parser_result.errorColumn();
```
yields
```
syntax error, unexpected STRING, expecting end of file at column 75
```

Issue #178, opened 2021-07-13
Support for Date/Time Operations


mweisgut:
Date/time operations are not sufficiently supported yet. Statements such as `"select (cast('2002-5-01' as date) + interval '60 days');"` cannot be parsed.

Such date/time operations are present in some TPC-DS queries, such as query 94.

[Postgres reference: Date/Time Functions and Operators](https://www.postgresql.org/docs/current/functions-datetime.html)

PR #190, opened 2021-09-10
Add Date Intervals

dey4ss:
This PR adds SQL date intervals, i.e., supports statemants such the following:
 ```SQL
 SELECT * FROM t 
    WHERE t.date BETWEEN '2021-01-01' AND CAST('2021-01-01' AS DATE) + 30 DAYS
```
This is achieved by a new type of `Expr`, with type `kExpressionInterval`. It is returned as second operand of the binary plus expression, where the start date ist the first one.

### CODE REVIEW

**Comment by dey4ss** [comment]
Regarding the syntax, `cast(... as date) + INTERVAL n days` is kind of a random: I've seen `cast(... as date) + n days` [[1]](https://github.com/gregrahn/tpcds-kit/blob/master/query_templates/query12.tpl) (which is implemented as well), `cast(... as date) + interval 'n' days` [[2]](https://github.com/Agirish/tpcds/blob/1d6a5d0461e379f425bb11df43a9bdd89f6dab75/modified/query12.sql), and Postgres supposes `cast(.. as date) interval 'n days'`[[3]](https://www.postgresql.org/docs/9.1/functions-datetime.html). However, I thougt this to be natural. If there are opinions preferring any of the variations, I'm open to it.

File: src/parser/flex_lexer.l [hunk 1]
```diff
@@ -192,12 +192,20 @@ HOUR        TOKEN(HOUR)
...
+MONTHS      TOKEN(MONTHS)
+YEARS       TOKEN(YEARS)
 TRUE		TOKEN(TRUE)
 FALSE		TOKEN(FALSE)
 TRANSACTION TOKEN(TRANSACTION)
 BEGIN       TOKEN(BEGIN)
 ROLLBACK    TOKEN(ROLLBACK)
 COMMIT      TOKEN(COMMIT)
+INTERVAL    TOKEN(INTERVAL)
+DAYS        TOKEN(DAYS)
```
mweisgut, line 208: Duplicates of `DAYS`?

dey4ss: good catch

dey4ss:
After revisiting the SQL standard and validation with Postgres, both `INTERVAL 'n' days` and `INTERVAL 'n days'` seems to be correct (but not `INTERVAL n days`. I will change the current behavior to apply this.

mweisgut:
I will do the review on Thursday at the latest.

src/parser/bison_parser.y
```diff
@@ -914,6 +915,7 @@ operand:
 	|	extract_expr
 	|	cast_expr
 	|	array_expr
+	|	interval_expression
```
klauck, line 918: I would suggest to introduce an 'inverval_literal' in ll. 1029 ('literal: [...]') instead. (interval_expression is used in the SQL standard to enforce type compatibility (see http://www.contrib.andrew.cmu.edu/~shadow/sql/sql1992.txt  p.168 ), which we do not do in the hyrise parser)

klauck:
> After revisiting the SQL standard and validation with Postgres, both `INTERVAL 'n' days` and `INTERVAL 'n days'` seems to be correct (but not `INTERVAL n days`. I will change the current behavior to apply this.

Arrg, the interval syntax is complex and differs depending on the specific database system: Further, note, it is `'n' day`, `'1 day'`, and `'5 days'` (with and without 's').

src/parser/bison_parser.y
dey4ss: I#m just about to change that. I guess Expr::isLiteral()`` should return true for interval literals, as well? And is there a reason not to include date literals here?

dey4ss: For now, I added `isType(kExprLiteralDate) || isType(kExprLiteralInterval)` to `Expr::isLiteral()`.

dey4ss:
>Further, note, it is 'n' day, '1 day', and '5 days' (with and without 's').
This ambiguation works now.

mweisgut (commented):
In addition to the above comments, it would be helpful if the description of this PR summarizes the different ways to define the intervals, i.e., `'n' day`, `'1 day'`, and `'5 days'`.

src/parser/bison_parser.y
```diff
@@ -1070,6 +1086,49 @@ date_literal:
 		}
 	;
 
+interval_literal:
+		int_literal duration_field { $$ = Expr::makeIntervalLiteral($1->ival, $2); delete $1; }
+	|   INTERVAL STRING opt_datetime_field {
```
klauck, line 1091: I would split the logic here und use
`|   INTERVAL STRING datetime_field
| INTERVAL STRING`

test/queries/queries-bad.sql
```diff
@@ -27,3 +27,14 @@
...
+!SELECT * FROM t WHERE a = DATE '2000-01-01' + INTERVAL 30 DAYS;
+!SELECT * FROM t WHERE a = DATE '2000-01-01' + INTERVAL 30 'DAYS';
+!SELECT * FROM t WHERE a = DATE '2000-01-01' + INTERVAL 'DAYS';
+!SELECT * FROM t WHERE a = DATE '2000-01-01' + INTERVAL '1' ANYTHING;
+!SELECT * FROM t WHERE a = DATE '2000-01-01' + INTERVAL '1 DAY' DAY;
+!SELECT * FROM t WHERE a = DATE '2000-01-01' + INTERVAL '30 ANYTHING';
+!SELECT * FROM t WHERE a = DATE '2000-01-01' + INTERVAL '30' DAYS;
+!SELECT * FROM t WHERE a = DATE '2000-01-01' + x DAYS;
+!SELECT * FROM t WHERE a = DATE '2000-01-01' + INTERVAL 'x' DAY;
+!SELECT * FROM t WHERE a = DATE '2000-01-01' + INTERVAL '3.3 DAYS';
```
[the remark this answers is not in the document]
dey4ss: The idea was to pass such checks to Hyrise. Also ensuring that intervals may only be input to + or - operations etc. would be out of the parser's scope for now. Regarding semantic sense, DB fiddle suggests that Postgres has no problem with Intervals in between Statements, e.g., `SELECT INTERVAL '2 days' BETWEEN INTERVAL '1 day' AND INTERVAL '3 days'`evaluates to 'true'.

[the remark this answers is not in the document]
mweisgut: Agreed. @klauck and I also discussed that and we also think that such semantic checks should be part of Hyrise.

https://sqliteonline.com/ also suggests that Postgres returns `true` for `SELECT INTERVAL '2 days' BETWEEN INTERVAL '1 day' AND INTERVAL '3 days'`.

mweisgut:
Referencing #184 and #178.

src/parser/flex_lexer.l
```diff
@@ -186,18 +186,25 @@ OF			TOKEN(OF)
...
-SECOND      TOKEN(SECOND)
-MINUTE      TOKEN(MINUTE)
-HOUR        TOKEN(HOUR)
-DAY         TOKEN(DAY)
-MONTH       TOKEN(MONTH)
-YEAR        TOKEN(YEAR)
+SECOND		TOKEN(SECOND)
+MINUTE		TOKEN(MINUTE)
+HOUR		TOKEN(HOUR)
+DAY			TOKEN(DAY)
```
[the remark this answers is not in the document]
dey4ss: The intendation in `bison_parser.y`  and `flexer_lexer.l` is kind of messy at all (mostly tabs, but sometimes four spaces). Indeed, tabs here (and for some of the tokens, spaces and tabs are mixed) look kind of weird when printed to console, but when they are fixed to contain tabs only (e.g., with nano), the intendation is not correct in a text editor (e.g., Sublime) anymore.
I would have no problem to fix this forever by applying a space intendation everywhere.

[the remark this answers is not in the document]
dey4ss: For reference, see the same lines of `flexer_lexer.l` opened with nano and Sublime:
image
image

src/parser/bison_parser.y
```diff
@@ -1070,6 +1086,49 @@ date_literal:
...
+			int duration{0}, chars_parsed{0};
+			char unit_string[8];
```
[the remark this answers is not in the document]
dey4ss: "'Seconds' and 'minutes' are the longest accepted interval qualifiers (7 chars) + null byte" ?

src/util/sqlhelper.cpp
```diff
@@ -93,7 +93,8 @@ namespace hsql {
     printExpression(expr->expr, numIndent + 1);
     if (expr->expr2 != nullptr) {
         printExpression(expr->expr2, numIndent + 1);
-    } else if (expr->exprList != nullptr) {
+    }
```
dey4ss, line 96: I changed this because I saw that a CASE statement only printed its 'else' expression and not the cases itself when fiddling. After having a look at #179 and #113, I'm not sure to keep it.

src/parser/flex_lexer.l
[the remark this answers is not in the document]
mweisgut: Having space indentation consistently would be the way to go in my opinion. However, the indentation you used in your added code looks consistent (tab indentation). For this PR, fixing the indentation issue is not required imo since you did not introduce it with this PR. However, feel free to fix it, if you want to. :)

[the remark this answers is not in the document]
dey4ss: I think openening a new PR for this once this one is merged is the cleanest solution.

src/util/sqlhelper.cpp
mweisgut: Similar to #179, I think we should pay attention to the test coverage for new code and add tests correspondingly. Since #113 has not been addressed yet, I suggest undoing the change and creating an issue for the mentioned CASE statement observation.

src/parser/bison_parser.y
```diff
@@ -1070,6 +1086,52 @@ date_literal:
...
+				free($2);
+				yyerror(&yyloc, result, scanner, "Found incorrect interval format. Expected format: INTEGER");
+				YYERROR;
+			}
+			free($2);
+			$$ = Expr::makeIntervalLiteral(duration, $3);
+		}
+	|   INTERVAL STRING {
+			int duration{0}, chars_parsed{0};
+			// 'Seconds' and 'minutes' are the longest accepted interval qualifiers (7 chars) + null byte
```
mweisgut, line 1104: Based on the string comparison cases below, it should be 'seconds' rather than 'Seconds', shouldn't it?

src/util/sqlhelper.cpp
dey4ss: I created #191 to address this.

src/parser/bison_parser.y
```diff
@@ -1070,6 +1086,52 @@ date_literal:
...
+			} else if (strcasecmp(unit_string, "day") == 0 || strcasecmp(unit_string, "days") == 0) {
+				unit = kDatetimeDay;
+			} else if (strcasecmp(unit_string, "month") == 0 || strcasecmp(unit_string, "months") == 0) {
+				unit = kDatetimeMonth;
+			} else if (strcasecmp(unit_string, "year") == 0 || strcasecmp(unit_string, "years") == 0) {
+				unit = kDatetimeYear;
+			} else {
+				yyerror(&yyloc, result, scanner, "Interval qualifier is unknown.");
+				YYERROR;
+			}
```
mweisgut, line 1130: The various cases should be coved by the test cases, i.e., every possible interval qualifier should be tested. Even though the CI pipeline does not include a test coverage check, line coverage should be our minimum test coverage criteria for new code IMO.

For checking the various interval qualifiers, checking the `datetimeField`, `ival`, and `type` variables should be fine since we already have test cases with more comprehensive checks.

src/parser/bison_parser.y
```diff
@@ -1007,6 +1008,20 @@ datetime_field:
...
+    |   MINUTES { $$ = kDatetimeMinute; }
+    |   HOURS { $$ = kDatetimeHour; }
+    |   DAYS { $$ = kDatetimeDay; }
+    |   MONTHS { $$ = kDatetimeMonth; }
+    |   YEARS { $$ = kDatetimeYear; }
+    ;
+
+duration_field:
+		datetime_field
+	|	datetime_field_plural
```
mweisgut, line 1022: Similar to the test coverage comment regarding the different unit strings, we should have at least a small interval test for each possible duration field.

src/parser/bison_parser.y
dey4ss: Is it sufficient to use a loop in the test, as I did for now?

src/util/sqlhelper.cpp
mweisgut: Thank you

src/parser/bison_parser.y
mweisgut: It's sufficient IMO

mweisgut:
Looks good to me. Just one minor change request:
Could you please add `SELECT (CAST('2002-5-01' as DATE) + INTERVAL '60 days');`, which is the example query of #178 to `queries-good.sql`? Then, we can close the issue with this PR IMO.

@klauck could you take a final look at this PR?

mweisgut (approved):

src/sql/Expr.cpp
```diff
@@ -140,6 +140,14 @@ Expr* Expr::makeDateLiteral(char* string) {
     return e;
 }
 
+Expr* Expr::makeIntervalLiteral(int64_t duration, DatetimeField unit) {
+    Expr* e = new Expr(kExprLiteralInterval);
+    e->name = strdup("INTERVAL");
```
klauck, line 145: I am not sure about the use of the `name` field here (and suggest removing the line (and also in `makeExtract` and `makeCast`)). I think the purpose of the `name` field was storing actual string values, which may differ among different Expression instances (e.g. different string literals) and not the name of the expression type.

klauck: Otherwise, all changes look fine for me. Thanks for your revision.

PR #199, opened 2022-01-07
Changed vector of SetOperations to SetOperation pointer

mweisgut:


mweisgut:
We should add test cases with nested SQL statements where a nested statement contains set operation(s).

klauck:
I think, we need the vector. (see my comments at #138)

PR #207, opened 2022-05-05
Add BIGINT DataType

dey4ss:
While playing around with the [Public BI Benchmark](https://github.com/cwida/public_bi_benchmark), I noticed that many queries and schemas use `BIGINT` as a data type, which is an 8-byte integer (cf. [Postgres](https://www.postgresql.org/docs/14/datatype-numeric.html)). Thus, `BIGINT` is equivalent to [our `LONG` type](https://github.com/hyrise/hyrise/blob/377befb4a05d4b5c866f8c19b327b99f300aebf9/src/lib/all_type_variant.hpp#L37)  and can be implemented either as own type or as an alias that is parsed to `DataType::Long`. 

This PR adds `BIGINT` as an own `DataType` (subject to discussion). After this PR is merged, we can easily update Hyrise to handle `BIGINT` as `LONG` values.


klauck:
Good catch.
Could you please add tests:

dey4ss:
I'm gonna add corresponding tests in the next days. Do you prefer an own `DataType` for `BIGINT` or should we only add the token and parse it to `LONG`?

klauck:
I would (leave it as it is) and add an own `DataType:BIGINT`.

klauck (approved):
Looks good to me.

dey4ss:
Oh, just happened to me accidentially while merging: Don't we have squash commits as default?

Issue #55, opened 2017-07-26
Improve memory management


torpedro:
Currently the parse tree contains a large number of raw pointers to maintain the nodes. This means that we always need to explicitly delete all children upon deletion of a node. While we do have an automatic memory leak checker, the risk of memory leaks is still there and handling memory this way is cumbersome. 

Possible improvements:
 * Use `std::string` instead of `char*`
 * Use smart pointers instead of raw pointers

We should measure the performance impact of using smart pointers. In general, I think the advantages of smart memory management will outweigh possible performance drawbacks.

Related Blog Post: [Feeding a Bison with tasty C++11 grAST!](https://coldfix.de/2015/05/16/bison-c++11/)

PR #210, opened 2022-05-09
Change Default Column Nullability

### LINKED ISSUE #209: Default nullability of columns
Currently, when we parse a `CREATE TABLE` statement, columns are only nullable if explicitly stated. See the following example:

```SQL
CREATE TABLE foo (
    a INT, -- not nullable
    b INT NULL, -- nullable
    c INT NOT NULL -- not nullable
);
```

This behavior differs from Postgres [1]:
> `NOT NULL`
    The column is not allowed to contain null values.
>
> `NULL`
    The column is allowed to contain null values. This is the default.
    This clause is only provided for compatibility with non-standard SQL databases. Its use is discouraged in new applications.

If we want to conform to Postgres, I would set the default nullability to true. I came across this due to loading tables from the Public BI Benchmark, which uses schemas in the way Postgres inteprets nullability. Loading the CSV files resulted in errors.

SQL-92 standard [2]:
>Every column has a nullability characteristic of known not nullable or possibly nullable, defined as follows:
>
>A column has a nullability characteristic that indicates whether any attempt to store a null value into that column will inevitably raise an exception, or whether any attempt to retrieve a value from that column can ever result in a null value. A column C with `<column name>` CN of a base table T has a nullability characteristic that is known not nullable if and only if either:
>
>   -  there exists at least one constraint that is not deferrable and that simply contains a `<search condition>` that contains CN IS NOT NULL or NOT CN IS NULL or RVE IS NOT NULL, where RVE is a `<row value constructor>` that contains a `<row value constructor expression>` that is simply CN without an intervening `<search condition>` that specifies OR and without an intervening `<boolean factor>` that specifies NOT.
>
>   -  C is based on a domain that has a domain constraint that is not deferrable and that simply contains a `<search condition>` that contains VALUE IS NOT NULL or NOT VALUE IS NULL without an intervening `<search condition>` that specifies OR and without an intervening `<boolean factor>` that specifies NOT.
>
>   -  CN is contained in a non-deferrable `<unique constraint definition>` whose `<unique specification>` specifies PRIMARY KEY.
>
>Otherwise, a column C is possibly nullable.

[1] https://www.postgresql.org/docs/14/sql-createtable.html
[2] https://www.contrib.andrew.cmu.edu/~shadow/sql/sql1992.txt

dey4ss:
This PR changes columns defined in `CREATE TABLE` statements to be nullable per default. With this change, we are compliant to Postgres [1] and the SQL standard [2].

Closes #209, where conflicts with Postgres and the SQL standard are explained. Additionally, this PR adds a test that permits primary keys to be nullable, i.e.,`CREATE TABLE a_table (a_column PRIMARY KEY NULL)` is invalid. Furthermore, it ensures ensure that the `PRIMARY KEY` constraint sets nullability to false.

[1] https://www.postgresql.org/docs/14/sql-createtable.html
[2] https://www.contrib.andrew.cmu.edu/~shadow/sql/sql1992.txt


klauck:
I can review this pull request on Wednesday.

dey4ss:
No pressure, I guess these minor changes do not have a high priority

dey4ss:
I'm not sure whether a `PRIMARY KEY` table constraint (not column constraint)  should ensure that the corresponding columns are not nullable as well (should be done [here](https://github.com/hyrise/sql-parser/blob/0c655f3a2981e2fd9258bd45c85411bcdf82e086/src/sql/CreateStatement.h#L100)). However, this would mean that we need to track whether a column is nullable as per default or explicitly requested to be nullable in to detect possible mismatches.

dey4ss:
Interesting: when building Hyrise, clang requires a `[[fallthrough]]` in the `CreateStatement`'s switch, but the CI's gcc does not like it. Building Hyrise with gcc@9 (sidon) is no problem. Maybe we should use a newer gcc version in the CI, the current seems to be 5 or 6.

klauck:
> This could be done via an additional flag or by using a dedicated `Nullable`enum (`Default`/`Yes`, `ExplicitYes`, `(Explicit)No`) as indicator for column nullability.

We could also keep the information in `column_constraints` and do not delete them in `(try)setNullableExplicit()`. This approach is what I would prefer: Parse the column definition and check for potential errors, but keep all information/elements in `column_constraints`. Then delete all unnecessary information at the end of `setColumnDefsAndConstraints()`.

klauck:
> However, the latter can also be ensured by the DBMS.

How many checks we want to do in the parser is up to us. There is a lot what can be done, but it takes time ;).

src/sql/CreateStatement.h
```diff
@@ -33,17 +33,42 @@ struct ColumnDefinition : TableElement {
 
   ~ColumnDefinition() override;
 
-  void setNullableExplicit() {
-    nullable = false;
+  bool trySetNullableExplicit() {
```
klauck, line 36: I personally do not like that we have to maintain state in the loop. (The old code was easier to read for me). I would simply go over the vector multiple times if the code is in turn easier to understand: Basically: if (NULL in vector and NOT NULL in vector): error.

src/sql/CreateStatement.h
```diff
@@ -33,17 +33,42 @@ struct ColumnDefinition : TableElement {
...
+          }
+          explicit_not_nullable = true;
+          nullable = false;
+          break;
+        default:
+          break;
       }
     }
+
+    for (auto constraint_index = constraints_to_remove.rbegin(); constraint_index != constraints_to_remove.rend();
```
klauck, line 66: We may want to keep this information and delete it later. See my overall comment.

dey4ss:
I just looked into Hyrise: we do not use any constraints (both table and column) and only consider a column's nullability. As we have information about a column being unique or primary key, we definitely should add this to Hyrise. Doing so, we can also check that table primary key constraints do not point to intendedly nullable columns. 
Altogether, this matches well with a set of column constraints where we do not delete explicitly requested nullability information (basically what we discussed offline yesterday).

src/sql/CreateStatement.h
dey4ss: I do not delete the information anymore. This way, the DBMS can check if a column is explicitly set to be nullable for a PRIMARY KEY table constraint.

dey4ss:
Still, the GCC update to GCC-9 is part of this PR, whereas it should not be necessary anymore. Do we want to keep it or maybe add an old (6?) and a new GCC (10+) version to the CI, maybe in an own PR?

Bouncner:
> Still, the GCC update to GCC-9 is part of this PR, whereas it should not be necessary anymore. Do we want to keep it or add an old (6?) and a new GCC (10+) version to the CI, maybe in an own PR?

I we can support GCC 6 without hazzles, I'd vote for keeping it. Adding GCC 10/11 is helpful nonetheless.

dey4ss:
> I we can support GCC 6 without hazzles, I'd vote for keeping it. Adding GCC 10/11 is helpful nonetheless.

I'd open another PR for this and reset gcc to 6 here

dey4ss:
I opened a PR for multiple compiler versions in the CI: #212.

src/parser/bison_parser.y
```diff
@@ -144,6 +144,7 @@
 
   std::vector<char*>* str_vec;
   std::vector<hsql::ConstraintType>* column_constraint_vec;
```
klauck, line 146: column_constraint_vec should be removed, because it is replaced by column_constraint_set, right?

dey4ss: absolutely

src/sql/CreateStatement.h
```diff
@@ -29,24 +30,31 @@ struct TableConstraint : TableElement {
...
-    for (unsigned long constraint_index = 0; constraint_index < column_constraints->size(); constraint_index++) {
-      if (column_constraints->at(constraint_index) == ConstraintType::Null) {
-        nullable = true;
-        column_constraints->erase(column_constraints->cbegin() + constraint_index);
-      } else if (column_constraints->at(constraint_index) == ConstraintType::NotNull) {
-        column_constraints->erase(column_constraints->cbegin() + constraint_index);
+  // By default, columns are nullable. However, we track if a column is explicitly requested to be nullable to
+  // notice conflicts with PRIMARY KEY table constraints.
+  bool trySetNullableExplicit() {
+    for (const auto& constraint : *column_constraints) {
```
klauck, line 40: I suggest to not iterate over all constraints, because we have a map.
```if (column_constraints->count(ConstraintType::NotNull) || column_constraints->count(ConstraintType::PrimaryKey) {
    if (column_constraints->count(ConstraintType::Null) {
        return false;
    }
    nullable = false;
}
return true;
```

klauck: I am unsure about the best way to check whether an element is in a map: `contains` is not supported by older compilers, right? It remains `find` vs. `count`.

klauck (changes requested):
Please, see individual comments. Besides, I am happy.

src/sql/CreateStatement.h
dey4ss: Exactly, `contains` is C++20. I used your suggestion with `count`

src/parser/bison_parser.y
```diff
@@ -508,6 +508,10 @@ create_statement : CREATE TABLE opt_not_exists table_name FROM IDENTIFIER FILE f
   $$->tableName = $4.name;
   $$->setColumnDefsAndConstraints($6);
   delete $6;
+  if (result->errorMsg()) {
+    delete $$;
+    YYERROR;
+  }
```
mweisgut, line 514: Isn't the `%destructor { delete ($$); } <*>` sufficient here for deallocating the memory?

dey4ss: If we omit L512, the memory leak check fails.

mweisgut: Okay. Seeing this reminds me of #55 and that addressing it might clean up the code.

mweisgut (approved):

klauck (approved):

klauck:
Thanks for your work, @dey4ss 👍

PR #212, opened 2022-05-18
Use Multiple Compiler Versions in CI

dey4ss:


.travis.yml
```diff
@@ -1,39 +1,48 @@
...
-  - tar -xvzf flex-2.6.4.tar.gz
-  - cd flex-2.6.4
-  - ./configure && make && sudo make install
-  - cd ..
-
-  # Show installed versions.
-  - which g++
-  - g++ -v
-  - bison --version
-  - flex --version
```
Bouncner, line 30: Can we still print these versions? Just to make it simpler to track which bison/flex etc. versions are provided by Ubuntu 16/20/..

PR #213, opened 2022-05-30
support for negative number

JoyBinY:


dey4ss:
Could you please provide some context for the adaption? In most cases, negative numbers should be treated as `unary minus` plus `unsigned integer`, resulting in the correct result. Is there a scenario we are not covering yet? By solely changing the lexer rule, I'm afraid that we end up enabling negative numbers for, e.g., `VARCHAR/CHARACTER VARYING`.

JoyBinY:
hi, and how to represent a negative number ?
`insert into persons(id, name, age, money) values (-200, 'write', -17.8, -23.45);`
the sql like this is not correct .
please show me the right sql, thank you !

JoyBinY:
we develop our database system base on sql-parser, and we do not know how to insert a negative number, so we edit the lexer rule.
on the current situation, is there any ways to add negative numbers ?

klauck:
@JoyBinY I simple fix, which is , however, not 100% correct is changing
-insert_statement : INSERT INTO table_name opt_column_list VALUES '(' literal_list ')' {
+insert_statement : INSERT INTO table_name opt_column_list VALUES '(' expr_list ')' {
in /src/parser/bison_parser.y line 675.
Without changing the parser, you could try another `insert` syntax: `insert into t select -1, -4, 'str' from dummy;`

Otherwise, we will update the parser within the next weeks (see #213)

Stephen-Pu:
Hello, @JoyBinY, I noticed that our fix for handling negative values were completed at the end of 2022, which means they were finished two years ago. However, this fix has not been merged into the main branch. 
What are the specific reasons for this? Since handling negative values is quite common in everyday use cases, not merging these changes could significantly impact the practical use of the code for users.

dey4ss:
Hi @Stephen-Pu, as is discussed in #215, the proposed solution in this PR solves the described issue. However, it allows negative numbers _everywhere_, including the cases where unsigned numbers are required. Thus, we do not consider the proposed patch as an option we want to merge.

As @klauck mentioned earlier, adapting the insert statement grammar rather than the lexer is our preffered solution, where negative numbers are treated as a unary minus expression containing a numerical integral. However, we did not pursue on that for capacity and priority reasons, as well as turnovers.

Because this PR has ben stale for a long time now, I will close it and have a look at #215 again.

PR #216, opened 2022-07-21
Update Toolchain for ARM Macs and add macOS CI stage

dey4ss:
This PR updates the Makefile and the format script to support the different brew library path on ARM Macs. Furthermore, it adds a macOS stage to the CI pipeline.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,8 +16,7 @@ matrix:
             - bison
             - flex
       env:
-        - MATRIX_EVAL="CC=gcc-6 && CXX=g++-6"
-
+        - MATRIX_EVAL="CC=gcc-6 && CXX=g++-6 && BISON=bison && FLEX=flex"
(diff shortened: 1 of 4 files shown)


Bouncner:
Do you think it makes sense to add a osx worker to travis?

dey4ss:
I think there's no reason not do do so. I'm gonna have a look at it.

dey4ss:
@Bouncner after quite some fiddling around, I added a macOS stage. To successfully run the tests, I suppressed memory leak checks when the tests do not run on a Linux machine (valgrind is only available there). Notably, the new CI stage uses Apple's clang rather than the brew version because installing llvm took quite long (up to 10 minutes, whereas the other stages run about 1-2 min). To still have get an indight on the used compiler, I print the compiler version in the CI.

Bouncner:
Awesome!

Bouncner (approved):
Awesome!

PR #217, opened 2022-08-03
Fix the `'dynamic_cast' not permitted with '-fno-rtti'` error when integrated with LLVM/MLIR.

septicmk:
A compile error occurs when including the `"SQLParser.h"` in another project. (sql_parser as third-party)

```
/thirdparty/sql-parser/src/sql/CreateStatement.h:74:67: error: 'dynamic_cast' not permitted with '-fno-rtti'
   74 | ynamic_cast<ColumnDefinition*>(tableElem)) {
      |                                         ^

/thirdparty/sql-parser/src/sql/CreateStatement.h:76:82: error: 'dynamic_cast' not permitted with '-fno-rtti'
   76 | dynamic_cast<TableConstraint*>(tableElem)) {
      |                                         ^
```

Separate the declaration and definition of the `setColumnDefsAndConstraints` should fix this.

### DIFF
--- a/src/sql/CreateStatement.cpp
+++ b/src/sql/CreateStatement.cpp
@@ -1,0 +1,19 @@
+
+#include "CreateStatement.h"
+
+namespace hsql {
+
+void CreateStatement::setColumnDefsAndConstraints(std::vector<TableElement*>* tableElements) {
+  columns = new std::vector<ColumnDefinition*>();
+  tableConstraints = new std::vector<TableConstraint*>();
+
+  for (auto tableElem : *tableElements) {
+    if (auto* colDef = dynamic_cast<ColumnDefinition*>(tableElem)) {
+      columns->emplace_back(colDef);
+    } else if (auto* tableConstraint = dynamic_cast<TableConstraint*>(tableElem)) {
+      tableConstraints->emplace_back(tableConstraint);
+    }
+  }
+}
+
+}  // namespace hsql
--- a/src/sql/CreateStatement.cpp
+++ b/src/sql/CreateStatement.cpp
@@ -1,4 +1,3 @@
-
 #include "SelectStatement.h"
 #include "CreateStatement.h"
 
--- a/src/sql/CreateStatement.cpp
+++ b/src/sql/CreateStatement.cpp
@@ -1,7 +1,59 @@
 
+#include "SelectStatement.h"
 #include "CreateStatement.h"
 
 namespace hsql {
+
+// CreateStatemnet
+CreateStatement::CreateStatement(CreateType type)
+    : SQLStatement(kStmtCreate),
+      type(type),
+      ifNotExists(false),
+      filePath(nullptr),
+      schema(nullptr),
+      tableName(nullptr),
+      indexName(nullptr),
+      indexColumns(nullptr),
+      columns(nullptr),
+      tableConstraints(nullptr),
+      viewColumns(nullptr),
+      select(nullptr){};
+
+CreateStatement::~CreateStatement() {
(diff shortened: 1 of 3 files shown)


Bouncner:
Any chance to can provide an example of how to trigger this issue?

Which clang version do you use?

septicmk:
The definition of `setColumnDefsAndConstraints` is in header file, and it tires to call [dynamic_cast](https://github.com/hyrise/sql-parser/blob/master/src/sql/CreateStatement.h#L74).
LLVM/MLIR project [implicitly add](https://github.com/llvm/llvm-project/blob/7f9e6f6fa6547b9c5648d96fd96920467c8b3e84/llvm/cmake/modules/AddLLVM.cmake#L47) `-fno-rtti` flag  (this flag disable `dynamic_cast`) and this will cause a compilation error as shown in the title.

Add `-fno-rtti` in example/Makefile will be an MRE.
```Makefile
#example/Makefile
CFLAGS = -std=c++1z -lstdc++ -Wall -fno-rtti -Werror -I../src/ -L../
all:
     $(CXX) $(CFLAGS) example.cpp -o example -lsqlparser

```

septicmk:
Hi @Bouncner, any chance to get this landed? This fix does not introduce any changes but only removes the function definition from the header file.

Bouncner (changes requested):
We just had our weekly discussion today. We certainly welcome your PR.

One small thing.
In the long run, we would like to remove the cluttered `statements.cpp`. For that, could you please move the function to a new `CreateStatement.cpp`?

septicmk:
> In the long run, we would like to remove the cluttered `statements.cpp`. For that, could you please move the function to a new `CreateStatement.cpp`?

Hi @Bouncner, the above issue has been addressed, PTAL.

src/sql/CreateStatement.cpp
```diff
@@ -0,0 +1,71 @@
+
+#include "SelectStatement.h"
```
Bouncner, line 2: Can you remove the first empty line?
And 2nd: do we need to include select statement here?

septicmk: In the destructor, we try to delete a SelectStatment. If not include the select statement here, the compiler complains:

```
src/sql/CreateStatement.cpp:25:10: error: possible problem detected in invocation of delete operator: [-Werror=delete-incomplete]
   25 |   delete select;
      |          ^~~~~~
src/sql/CreateStatement.cpp:25:10: error: invalid use of incomplete type 'struct hsql::SelectStatement' [-Werror]
In file included from src/sql/SQLStatement.h:6,
                 from src/sql/CreateStatement.h:5,
                 from src/sql/CreateStatement.cpp:1:
src/sql/Expr.h:10:8: note: forward declaration of 'struct hsql::SelectStatement'
   10 | struct SelectStatement;
      |        ^~~~~~~~~~~~~~~
src/sql/CreateStatement.cpp:25:10: note: neither the destructor nor the class-specific operator delete will be called, even if they are declared when the class is defined
   25 |   delete select;
      |          ^~~~~~
cc1plus: all warnings being treated as errors
```

septicmk: The first empty line has been removed.

Bouncner: Ah yeah, right. Got it.

Fine from my side. But since my knowledge about the parser is limited, I'd like to get another opinion on it.

dey4ss (approved):

Bouncner (approved):

Bouncner:
Feel free to merge.

PR #222, opened 2022-10-19
Remove contributors

Bouncner:
This PR removes the contributors list.
The list is anyways already listed by GitHub, showing who contributed how much to the parser.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,10 @@ matrix:
             - bison
             - flex
       env:
-        - MATRIX_EVAL="CC=gcc-6 && CXX=g++-6 && BISON=bison && FLEX=flex"
+        - CC=gcc-6
+        - CXX=g++-6
+        - BISON=bison
+        - FLEX=flex
 
     - name: ubuntu22.04-gcc-12
       os: linux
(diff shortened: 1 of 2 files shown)


Bouncner:
Running Travis on an old version of GCC should be sufficient. I don't see a value in listing support for a ten year old compiler. :)

dey4ss (approved):

dey4ss:
The macOS build was cancelled because brew upgarded lots of formulae. Can you try adding
`- export HOMEBREW_NO_AUTO_UPDATE=1 HOMEBREW_NO_INSTALL_CLEANUP=1`
for the macOS builds / in the `before_install` section of the `.travis.yml`? (see https://travis-ci.community/t/cannot-avoid-brew-update/12033/2)

dey4ss:
Hm, I'm not sure if Travis has hiccups or if I gloriously guided you to break the CI 😅
In the latter case, adding the environment variables here could be an alternative:
https://github.com/hyrise/sql-parser/blob/8a003c0c6af87d44aca50b5f824b948fca346d29/.travis.yml#L58
However, homebrew upgrading a bunch of irrelevant formulae seems to have been a Travis issue for quite some time and I wonder how we did not have more problems with that in the past.

PR #225, opened 2022-11-17
Enable Queries on Imported/Exported Tables

dey4ss:
In hyrise/hyrise#2514, we sketched how filtering and encoding selection during import and export of tables does look like in Posgtres [1]. After a discusison in the weekly, this PR enables filters on import/export of tables. The export filter will also be implemented in Hyrise.

[1] https://www.postgresql.org/docs/15/sql-copy.html


test/queries/queries-good.sql
```diff
@@ -70,11 +70,13 @@ COPY students FROM 'file_path' WITH FORMAT TBL;
 COPY students FROM 'file_path' WITH FORMAT CSV;
 COPY students FROM 'file_path' WITH FORMAT BIN;
 COPY students FROM 'file_path' WITH FORMAT BINARY;
+COPY good_students FROM 'file_path' WHERE grade > (SELECT AVG(grade) from alumni);
 COPY students TO 'student.tbl';
 COPY students TO 'file_path' WITH FORMAT TBL;
 COPY students TO 'file_path' WITH FORMAT CSV;
 COPY students TO 'file_path' WITH FORMAT BIN;
 COPY students TO 'file_path' WITH FORMAT BINARY;
+COPY (SELECT firstname, COUNT(*) FROM students GROUP BY firstname) TO 'student_names.csv';
```
Bouncner, line 79: Why not `file_path`?

Bouncner: Besides that nothing to add.

I am nonetheless wondering, if we have tested all possible (or at least relevant) cases. But tbh, I have used queries on exports and I wasn't even aware it existed.

I guess an empty results is just a CSV file (if requested) with a header, right?

Bouncner (commented):
I don't have much to say.

test/queries/queries-good.sql
dey4ss: Just something figurative here for the example query. Can change if requested.

I just hooked the existing select statement into the export statement: ﻿https://github.com/hyrise/sql-parser/blob/396fd59aa6bd154e330eb485fee1473245a6df84/src/parser/bison_parser.y#L475
The select statement should be tested accordingly in [`select_tests.cpp`](https://github.com/hyrise/sql-parser/blob/396fd59aa6bd154e330eb485fee1473245a6df84/test/select_tests.cpp). Same goes for the `WHERE`clause of the import statement.

Last question: The Hyrise exporter will write the header regardless of the content (so an empty table results in the written header - however, we do not write CSV headers but create an extra file containing the schema information). If the desired table does not exist, we fail.

Bouncner:
Fine from my side, but as I am not familiar with the parser, maybe wait for another approval?

src/sql/statements.cpp
```diff
@@ -259,39 +266,39 @@ SelectStatement::~SelectStatement() {
...
   }
 
-  if (lockings != nullptr) {
+  if (lockings) {
     for (LockingClause* lockingClause : *lockings) {
-      if (lockingClause->tables != nullptr) {
+      if (lockingClause->tables) {
         for (char* dtable : *lockingClause->tables) {
-          if (dtable != nullptr) free(dtable);
+          if (dtable) free(dtable);
```
klauck, line 301: Simply `free(dtable)`?

src/parser/bison_parser.y
```diff
@@ -36,8 +36,7 @@
 // Specify code that is included in the generated .h and .c files
 // clang-format off
 %code requires {
-  // clang-format on
```
klauck, line 39: Why remove? Does clang complain about the format?

src/sql/statements.cpp
```diff
@@ -343,15 +350,15 @@ TableRef::~TableRef() {
...
+  if (list) {
     for (TableRef* table : *list) {
       delete table;
     }
     delete list;
   }
 }
 
-bool TableRef::hasSchema() const { return schema != nullptr; }
+bool TableRef::hasSchema() const { return static_cast<bool>(schema); }
```
klauck, line 361: For me, the old version is clearer. But I'm not sure what is generally preferred.

src/sql/Expr.cpp
```diff
@@ -250,12 +250,12 @@ bool Expr::isLiteral() const {
          isType(kExprLiteralNull) || isType(kExprLiteralDate) || isType(kExprLiteralInterval);
 }
 
-bool Expr::hasAlias() const { return alias != nullptr; }
+bool Expr::hasAlias() const { return static_cast<bool>(alias); }
```
klauck, line 253: see below

src/sql/Expr.cpp
```diff
@@ -250,12 +250,12 @@ bool Expr::isLiteral() const {
...
 
-bool Expr::hasTable() const { return table != nullptr; }
+bool Expr::hasTable() const { return static_cast<bool>(table); }
```
klauck, line 255: see below

klauck (approved):
Looks good to me.

I had only
-   one question
-   two style questions (but the proposed versions are ok for me)

.

src/parser/bison_parser.y
dey4ss: To be honest, gonna have to look into this after that time. Think that clang-format at that time (probably 14 or so) did some weird stuff then.

src/sql/statements.cpp
dey4ss: To me, comparing a pointer `==`/`!=` `nullptr` doesn't seem right at all (not only in if clauses). Thus, the change. Maybe @Bouncner has an opinion on that

src/sql/statements.cpp
dey4ss: cppreference:
> If ptr is a null pointer, the function does nothing. 

Absolutely, gonna remove the mentioned ifs.

src/sql/statements.cpp
Bouncner: To me, comparing a pointer to `nullptr` is more obvious than casting it to another type which does implicitly the same.

dey4ss: Alright, I'm outvoted :D

src/parser/bison_parser.y
dey4ss: Indeed, clang-format (15) locally screws up the indentation and even adds linebreaks in the bison control lines.

dey4ss:
@Bouncner @klauck should be ready to go now. Knaller? 🧨 💥

Bouncner:
> @Bouncner @klauck should be ready to go now. Knaller? 🧨 💥

🧨 💥 ❗

Bouncner (approved):

dey4ss:
Argh, Travis doesn't wanna work again 😡 
Claims our credits are completely consumed while we still have lots of them, doesn't even list this branch as a PR. Starting to really hate that s***. Any ideas?

klauck (approved):

Bouncner:
Mailing TravisCI? :-/

dey4ss:
Yeah, already did that. Let's see what happens.

dey4ss:
They fixed it.

PR #230, opened 2023-03-17
[WIP] Use Gihub Actions as CI

dey4ss:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,86 @@
...
+            cxx: g++-6
+            os: ubuntu-latest
+            container: ubuntu:18.04
+
+          - name: gcc-12
+            cc: gcc-12
+            cxx: g++-12
+            os: ubuntu-latest
+            container: ubuntu:22.04
+            build_options: "relaxed_build=on"
```
Bouncner, line 35: Comment why?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,86 @@
...
+
+          - name: clang-15
+            cc: clang-15
+            cxx: clang++-15
+            os: ubuntu-latest
```
Bouncner, line 40: At some point, `ubuntu-latest` will differ from `container: ubuntu:22.04`. Does it make sense to list both?

dey4ss: You mean to always have a somewhat up-to-date Ubuntu in the CI? In this case, I would to not specify a clang version and add the following confoguration:
```yml
- name: clang-latest
  cc: clang
  cxx: clang
  os: ubuntu-latest
```

Bouncner: Not really what I meant. But that might be a good point. I think the idea is good with using a default "latest ubuntu and latest clang", BUT it might trigger problems that we don't have under control. I might have different opinion on Hyrise, but for the parser, I would be somewhat more conservative and be very specific about the versions.

What I really wanted to ask: what happens if the os with `ubuntu-latest` is going to be something "newer" like 23.10 and we still request `contrainer: ubuntu:22.04`. So we might be referencing two different versions of ubuntu in these two lines.

dey4ss: Ah, got you. Well, these lines do exactly what they are supposed to: They run on the `latest Ubuntu` runner and inside this runner, run a Docker container with `Ubuntu 22.04`. Of course that's useless now when `latest == 22.04`, but will run our suite on 22.04 till the end of days (or GH actions). If you look at the GCC-6 stage, you'll notice that it works there pretty well: We run on the latest runner, but our suite is run on Ubuntu 18.04 where we can acutally install GCC-6. (This design was inspired by [Boost](https://github.com/boostorg/integer/blob/8ce66ab37e246d1df51a8d7b37cfa0049057ca8f/.github/workflows/ci.yml#L36))

Bouncner (approved):
LGTM

PR #233, opened 2023-05-30
Add window functions

dey4ss:
This PR adds SQL:2003 window functions compatible to [Postgres](https://www.postgresql.org/docs/15/sql-expressions.html#SYNTAX-WINDOW-FUNCTIONS). The window is added as an aditional description for (function) expressions. It consists of a (possibly empty) list of Expressions as `PARTITION BY` expressions, a (possible empty) list of `ORDER BY` expressions, and a (possible default-initialized) frame definition.

In the current implementation, the frame definition (frame type and bounds) is realized as structs with enum values for the frame/bound type. Initially, I wanted to keep it simple and just pass strings, but this approach had several caveats: (i) the frame type cannot be an unquoted string (aka `IDENTIFIER`) since that causes a grammar error (shift conflict), and (ii) since frame bound specifiers consist of either two strings or unsigned int and string, merging them into a single string is cumbersome (requires `strcpy`, `strcat`, and `malloc` calls – we cannot even use `new` for `char*`), so I decided against this approach.

Note that `UNBOUND`, `PRECEDING`, and `FOLLOWING` are implemented as tokens. Thus, they are reserved from now on. Though Postgres [does not list them as reserved keywords](https://www.postgresql.org/docs/current/sql-keywords-appendix.html), I think we can live with this shortcoming. Otherwise, I can also implement a solution where we `strcasecmp` the frame types/bounds.

Files to look at are mainly `bison_parser.y` (actual new grammar), `flex_lexer.l` (new tokens), `Expr.*` (new Expression + helper stuctures), and `select_tests.cpp` (tests). `sqlhelper.*` has new functionality to print window functions, `queries-[good|bad].sql` contain more examples, and the rest of the code is auto-generated.

No pressure, but it would be great if we could merge this before the DYOD project start.


test/select_tests.cpp
```diff
@@ -988,3 +988,131 @@ TEST(MultipleLockingClause) {
...
+TEST(WindowExpression) {
+  SelectStatement* stmt;
+  TEST_PARSE_SQL_QUERY(
+      "SELECT t2, avg(t1) OVER() FROM t;"
+      "SELECT rank() OVER(ORDER BY t1) FROM t;"
+      "SELECT avg(t1) OVER(PARTITION BY t2, t3 ORDER BY t4, t5 ROWS UNBOUNDED PRECEDING) FROM t;"
+      "SELECT rank() OVER(PARTITION BY t1 ORDER BY t2 ROWS BETWEEN 25 PRECEDING AND 2 FOLLOWING) FROM t;"
+      "SELECT rank() OVER(PARTITION BY t1 ORDER BY t2 RANGE BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING) FROM "
+      "t;"
+      "SELECT rank() OVER(PARTITION BY t1 ORDER BY t2 GROUPS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW) FROM t;",
```
Bouncner, line 1001: Can I haz two windows?

Bouncner (approved):
Looks good to me, but due to my limited knowledge of the parser, I hope for another review.

dey4ss:
@klauck @Bouncner Windows are now limited to be defined for function expressions.

Bouncner:
The changes look good to me.

src/parser/bison_parser.y
```diff
@@ -978,8 +987,35 @@ comp_expr : operand '=' operand { $$ = Expr::makeOpBinary($1, kOpEquals, $3); }
 | operand LESSEQ operand { $$ = Expr::makeOpBinary($1, kOpLessEq, $3); }
 | operand GREATEREQ operand { $$ = Expr::makeOpBinary($1, kOpGreaterEq, $3); };
 
-function_expr : IDENTIFIER '(' ')' { $$ = Expr::makeFunctionRef($1, new std::vector<Expr*>(), false); }
-| IDENTIFIER '(' opt_distinct expr_list ')' { $$ = Expr::makeFunctionRef($1, $4, $3); };
+function_expr : IDENTIFIER '(' ')' opt_window { $$ = Expr::makeFunctionRef($1, new std::vector<Expr*>(), false, $4); }
```
klauck, line 990: I would split the general function expression (which can occur also in the where clause) and the specific window function, which is only allowed in the SELECT and ORDER BY clause and comes with a special syntax, requiring the keyword "OVER".

I think, this would, in particular, benefical for future extensions, such as advanced aggregate expression.

src/parser/bison_parser.y
```diff
@@ -978,8 +987,35 @@ comp_expr : operand '=' operand { $$ = Expr::makeOpBinary($1, kOpEquals, $3); }
...
+opt_window : OVER '(' opt_partition opt_order frame_clause ')' { $$ = new WindowDescription($3, $4, $5); }
+| /* empty */ { $$ = nullptr; };
+
+opt_partition : PARTITION BY expr_list { $$ = $3; }
+| /* empty */ { $$ = nullptr; };
+
+// We use the Postgres default if the frame end or the whole frame clause is omitted. "If `frame_end` is omitted, the
+// end defaults to `CURRENT ROW`. [...] The default framing option is `RANGE UNBOUNDED PRECEDING`, which is the same as
+// `RANGE BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW`."
+frame_clause : frame_type frame_bound { $$ = new FrameDescription{$1, $2, new FrameBound{0, kCurrentRow, false}}; }
```
klauck, line 1004: `opt_frame_clause` (naming) because the frame clasue is optional?

src/sql/Expr.cpp
```diff
@@ -5,6 +5,39 @@
...
+    : type{type}, start{start}, end{end} {}
+
+FrameDescription::~FrameDescription() {
+  delete start;
+  delete end;
+}
+
+WindowDescription::WindowDescription(std::vector<Expr*>* partitionList, std::vector<OrderDescription*>* orderList,
+                                     FrameDescription* frameDescription)
+    : partitionList{partitionList}, orderList{orderList}, frameDescription(frameDescription) {}
```
klauck, line 21: `frameDescription{frameDescription}` for a consistent style within the constructor.
I am not sure whether to use the newer `{}` syntax or not. (up to you)
The older `()` style is still dominant in the rest of the code.

src/parser/bison_parser.y
klauck: As discussed today, we could add a comment:
`// function_expr is used for window functions, aggregate expressions, and functions calls because we run into shift/reduce conflicts when splitting them`

klauck (approved):
Thanks again for your work! Feel free to merge the changes after checking my final comments.

PR #240, opened 2024-05-07
Correctly parse named columns for joins

### LINKED ISSUE #239: Segmentation fault when: [...] USING (*)
Hi,

## The bug
A statement like this `SELECT * FROM abc INNER JOIN xyz USING (*)` results in a segmentation fault. But `SELECT * FROM abc JOIN xyz USING (id)` parses successfully. The parser does not seem to like the `*` in the `USING`.

This was found using grammar-based fuzzing.

---

## How it was found, if you think that is fun :)

I am doing some analysis of grammar-based fuzzy testing for my master's thesis. For this I used AFL++ and created a grammar for SQL, which AFL++ then uses for its tests. Part of the grammar can be seen here. The highlights are a path to an SQL statement that crashes the parser:
image

(1) We start from `<START>` and go to `<STATEMENTS>`.
(2) Here it can only go to `<STATEMENT>` (singular)
(3) Here it has multiple options. For the crash-statement, it takes the `<SELECT>`
(4) Now it takes the `<SELECT>` with the `<JOIN_CLAUSE>`
(5) Here it takes the one with "USING"
(6) It takes the one with the `*` and goes to `<COLUMNS_1>`
(7) Here it takes the empty one

This could result in the following statement: `SELECT v AS Y6 FROM b INNER JOIN pB USING (* );`, which gives a segmentation fault.

dey4ss:
This PR allows a list of identifiers (a.k.a column names) when joins name columns with `USING (...)`. This behavior is in line with the SQL standard and PostgreSQL. Furthermore, we add a flag if join conditions come from specifying them directly (`foo JOIN bar ON (...)`) or from named columns (`foo JOIN bar USING (...)`) so the database can adjust the emitted columns addording to the SQL standard (emit both columns for each predicate in `ON`, emit a single column for each column in `USING`).

For details, see #239.

Fixes #239.


Bouncner:
What is your take on the “no column references” limitation?

dey4ss:
First of all, it is in line withe the standard and pgSQL, which we try to mostly comply to. Second, it also that makes sense IMO. `foo JOIN bar USING (...)` is comparable to `foo NATURAL JOIN bar` (the latter is equivalent to the first when the `USING` list contains all columns with the same name in `foo` and `bar` – `USING` is a `NATURAL` join where you explicitly specify or limit the used columns). All columns in the list have to be present in both `foo` and `bar`, so a column reference explicitly referencing a column on only one side is not really useful.

dey4ss:
Instead of parsing the list of columns as conjunctions of equality predicates and adding an extra boolean flag to `JoinDefinition`, we could also simply have the vector of column names as member:

```c++
struct JoinDefinition {
  JoinDefinition();
  virtual ~JoinDefinition();
  TableRef* left;
  TableRef* right;
  Expr* condition;
  std::vector<char*>* namedColumns;

  JoinType type;
};
```

This would be a bit more explicit and leaves resolving the column names to the DBMS. However, it differs to the way we handled `USING` before, and I am not sure what we want to do here in terms of backwards compatibility/divergence from forks.

test/select_tests.cpp
```diff
@@ -348,7 +348,8 @@ TEST(SelectJoin) {
   ASSERT_STREQ(inner_join->left->name, "fact");
   ASSERT_EQ(inner_join->right->type, kTableName);
   ASSERT_STREQ(inner_join->right->name, "City")
```
Bouncner, line 350: missing ;

Bouncner (approved):
Not much to say.

test/queries/queries-bad.sql
```diff
@@ -60,3 +60,10 @@
 !SELECT rank() OVER (ROWS UNBOUNDED PRECEDINGG) FROM test;
 !SELECT test1, rank() OVER (ROWS -1 PRECEDING) FROM test;
 !SELECT test1, rank() OVER (ROWS BETWEEN UNBOUNDED PRECEDING AND -1 FOLLOWING) FROM test;
+# Join USING
+!SELECT * FROM foo INNER JOIN bar USING ();
+!SELECT * FROM foo INNER JOIN bar USING (*);
+!SELECT * FROM foo INNER JOIN bar USING (foo.a);
```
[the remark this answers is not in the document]
dey4ss: https://github.com/hyrise/sql-parser/pull/240#issuecomment-2098690293

and the following answer to that.

test/select_tests.cpp
dey4ss: right. But it parses. Had a look into the microtest header and there, semicolons are already set (even superfluous ones: the basic asserts have semicolons already and the macros that build upon them add extra semicolons... 🥴). Is that something we want to touch? In the best case, this just means that we remove some semicolons in the microtest header.

dey4ss: https://github.com/hyrise/sql-parser/blob/c2471248cef8cd33081e698e8ac65d691283dbd4/test/thirdparty/microtest/microtest.h#L24-L41

Bouncner: Since we cannot touch the third party lib, I would care about our code and here missing semicolons bother me. :)

I don't mind if the compiled code ends up having a few unnecessary semicolons.

test/queries/queries-bad.sql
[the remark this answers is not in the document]
Bouncner: Can you add a comment then? People won't look up this PR when they stumble over this.

test/select_tests.cpp
dey4ss: Just adapted the header, can revert if you like. It helped spotting multiple places where assertions did not have semicolons at the end. IMO, if macros have semicolons internally, we should not add superfluous ones to make their calls more function-like.

Bouncner: Love it!

Bouncner:
Feel free to zünd.

PR #242, opened 2024-06-03
Allow more expressions for selected statements

### LINKED ISSUE #215: Correct Handling of Negative Numerical Literals
As mentioned by @JoyBinY in #213, the parser does not handle negative numerical literals correctly in all cases, especially for `INSERT` statements.

Code for Reproduction:
```SQL
INSERT INTO foo VALUES (1, 'bar', -2);
                                  ^=== ERROR HERE!     
```

An `INSERT INTO ... VALUES` statement has a literal list to represent the values. A literal is defined as 
```
literal : string_literal | bool_literal | num_literal | null_literal | date_literal | interval_literal | param_expr;
```

The numeric literal can be either an integer or a float (both parsed from _unsigned_ values by the lexer). We represent a negative number as unary `-` operator expression with the numeric literal as input in other cases, which is not a literal anymore.

However, the solution suggested in #213 is not sufficient either, as allowing negative values for all integers also enables, e.g., the following statements:
```SQL
CREATE TABLE foo (bar VARCHAR(-5));
SELECT  TOP -3 bar FROM foo;
```

A possible solution (option 1) could be to expand the `num_literal` definition from:
```
num_literal : FLOATVAL { $$ = Expr::makeLiteral($1); }
| int_literal;
```
to something like:
```
num_literal : FLOATVAL { $$ = Expr::makeLiteral($1); }
| int_literal;
| '-' FLOATVAL { $$ = Expr::makeOpUnary(kOpUnaryMinus, Expr::makeLiteral($2)); }
| '-' int_literal { $$ = Expr::makeOpUnary(kOpUnaryMinus, $2); };
```

or:
```
unsigned_num_literal: FLOATVAL { $$ = Expr::makeLiteral($1); }
| int_literal;

num_literal : unsigned_num_literal
| '-' unsigned_num_literal {
  $2->ival = $2->ival * -1;
  $2->fval = $2->fval * -1;
  $$ = $2;  // or just $$ = Expr::makeOpUnary(kOpUnaryMinus, $2); 
};
```

As the `num_literal`, in contrast to `[INT|FLOAT]VAL` and `int_literal`, is only used in the `literal` definition, we should not interfere with the definitions requesting unsigned integers.
An alternative (option 2) could be to define unsigned and signed versions of `[INT|FLOAT]VAL` in the lexer and use these versions accordingly in all definitions.
@Bouncner @klauck @mweisgut what are your opinions on this topic?

Furthermore, I noticed that `INSERT` statements are not properly tested in the [SQL tests](https://github.com/hyrise/sql-parser/blob/4381d544c1e2d97610cf3deb315f66eaaa39fcf9/test/sql_tests.cpp#L211), negative numbers in `INSERT` statement do not appear in our test queries [(`queries_good.sql`)](https://github.com/hyrise/sql-parser/blob/4381d544c1e2d97610cf3deb315f66eaaa39fcf9/test/queries/queries-good.sql), and neither do the permitted examples from above [(`queries_bad.sql`)](https://github.com/hyrise/sql-parser/blob/4381d544c1e2d97610cf3deb315f66eaaa39fcf9/test/queries/queries-bad.sql)

dey4ss:
As raised in #213 and discussed in #215, some valid expressions, such as negative numbers, are not supported by the `INSERT` statement. Thus, this PR carefully extends the range of expressions allowed in the `INSERT` statements, but also in `EXECUTE` statements and in hints.

I looked into SQL:2023 to see which expressions are allowed inside the `VALUES` list of an `INSERT` statement,  and there are a lot of valid alternatives in the standard. Some of them are collected in the following:

```diff
-| <scalar subquery>
-| <case expression>
+| <cast specification>
 | <boolean value expression>
 | <numeric value expression>
 | <string value expression>
 | <datetime value expression>
 | <interval value expression>
-| <user-defined type value expression>
-| <reference value expression>
-| <collection value expression>
-| <JSON value expression>
```

Note that all of this is scattered accross many pages and places as part of complex hierarchies. I collected just a few possibilities here. I marked newly supported (+) and some unsupported (-) alternatives. Furthermore, this PR also allows for _signed_ numerical/interval expressions (a.k.a unary minus of numeric/interval). Cast expressions are restricted to only take other allowed expressions as input.

Besides, this PR updates the CI runners to Ubuntu 24.04 and uses its shipped clang/gcc versions. While checking the runners, I noticed that (i) there was an issue with the member variable introduced in #240 and (ii) that did not cause the CI to fail. I sneaked in fixes for both.

fixes #215


sandrain:
Looking forward to seeing this patch land. 👍

Issue #243, opened 2024-06-21
Parser fails if a column is called "locked"


kingkero:
To demonstrate quickly, I just updated an existing test:

```diff
 TEST(SelectSchemaTest) {
-  TEST_PARSE_SINGLE_SQL("SELECT grade FROM some_schema.students;", kStmtSelect, SelectStatement, result, stmt);
+  TEST_PARSE_SINGLE_SQL("SELECT grade, locked FROM some_schema.students;", kStmtSelect, SelectStatement, result, stmt);

   ASSERT(stmt->fromTable);
   ASSERT_EQ(std::string(stmt->fromTable->schema), "some_schema");
 }
```

Results in:

```bash
{ running} SelectSchemaTest
{  failed} SelectSchemaTest
           Assertion failed: result.isValid()
           test//select_tests.cpp:158
```

With backticks I get

```bash
[SQL-Lexer-Error] Unknown Character: `
```

It seems that since MySQL 8.0.1 `LOCKED` is a [reserved keyword](https://dev.mysql.com/doc/refman/8.0/en/keywords.html), but even with with 8.0.27 I could easily use it without backticks in a query ([see fiddle](https://sqlfiddle.com/mysql/online-compiler?id=26c32048-81e7-48a4-8516-bf1058497501)).

Bouncner:
Since it is a keyword (even though non-reserved) in Postgres (which is the SQL dialect that are following), the error does not surprise me.
You should be able to normally quote the keyword and still use it as a columns name (I cannot test it right now). But ``` is MySQL-only and not following SQL standard. Can you try `"`?

klauck:
`"` and `'` work.
You can also use the `example` executable to quickly test parsing specific statements:

`./example/example 'SELECT grade, "locked" FROM some_schema.students;'`
Parsed successfully!
Number of statements: 1
SelectStatement
	Fields:
		grade
		locked
	Sources:
		students
			Schema
				some_schema

kingkero:
Thank you both for the quick replies!

Can confirm that it works with quotes (wasn't aware that backticks are MySQL-only)

PR #244, opened 2024-06-21
Add description for quickly running/testing custom SQL statements

klauck:
The possibility for quickly running/testing SQL statements with the parser is not documented in the README.

Some issues, such as #243, could, for example, described and tested with the command line executable `./example/example`.

This pull request documents this feature.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,18 @@ To use the SQL parser in your own projects you simply have to follow these few s
     }
 }
 ```
+  7. The corresponding `example` executable can be used to test/parse specific custom SQL statements:
+     
+     Running, for example, `./example/example "SELECT * FROM test;"` produces:
+```
+Parsed successfully!
+Number of statements: 1
+SelectStatement
+	Fields:
+		*
+	Sources:
+		test
+```     
 
 Quick Links:
 


klauck:
I am not sure where to place this documentation:

1. Under the example project (as is)
2. As a separate feature

1 makes sense, because compilation is also required and the code is referenced
2 makes sense, because the executable can be used to test statements without having an own project

Bouncner:
I was not even aware of the example binary. Can we find a better name? Repl? Console?

> I am not sure where to place this documentation:

What documentation do you mean exactly?

dey4ss:
> Can we find a better name? Repl? Console?

It's not interactive, you just provide the query as CLI argument and it prints it. I guess the name comes from the fact that it's basically a code snippet that shows how to use the parser in an application – with the side effect of printing the statement. But I don't have a better idea for a name

klauck:
> I was not even aware of the example binary

Yeah, and it is a great possibility to try out the parser. This is why I would like to add a documentation of this feature in the README.

> What documentation do you mean exactly?

The proposed change of the README. My question is how to integrate the change in the existing README.
The current proposal is to put it as point 7. under "To use the SQL parser in your own projects you simply have to follow these few steps." I think this place is not a perfect fit.

Maybe, It is better to separate the build and test process  1.-4. and then have the option to

1. Use it as a library

or 2. try it out running `./example/example`

README.md
```diff
@@ -13,37 +13,58 @@ In March 2015 we've also written a short paper outlining discussing some develop
 
 **Note:** You can also find a detailed usage description [here](docs/basic-usage.md).
 
-To use the SQL parser in your own projects you simply have to follow these few steps.
+
+### Build and Test
 
  1. Download the [latest release here](https://github.com/hyrise/sql-parser/releases)
  2. Compile the library `make` to create `libsqlparser.so`
  3. *(Optional, Recommended)* Run `make install` to copy the library to `/usr/local/lib/`
```
Bouncner, line 21: Not yours, but I wouldn't consider it "recommended" to install the library. I would be rather surprised if somebody installs it system-wide.

README.md
```diff
@@ -13,37 +13,58 @@ In March 2015 we've also written a short paper outlining discussing some develop
 
 **Note:** You can also find a detailed usage description [here](docs/basic-usage.md).
 
-To use the SQL parser in your own projects you simply have to follow these few steps.
+
+### Build and Test
 
  1. Download the [latest release here](https://github.com/hyrise/sql-parser/releases)
```
Bouncner, line 19: Well, well, well ...

image

README.md
```diff
@@ -13,37 +13,58 @@ In March 2015 we've also written a short paper outlining discussing some develop
...
  2. Compile the library `make` to create `libsqlparser.so`
```
Bouncner, line 20: the library USING `make` ?

Bouncner: `.so` is linux-only. Just say "to create the library?

README.md
```diff
@@ -13,37 +13,58 @@ In March 2015 we've also written a short paper outlining discussing some develop
...
+	
+	```cpp
+	#include "hsql/SQLParser.h"
+	
+	/* ... */
+	
+	{
+	    // Basic Usage Example
+	
+	    const std::string query = "...";
```
Bouncner, line 53: To annoy people, I would use auto-to-stick here. 🫣

README.md
```diff
@@ -13,37 +13,58 @@ In March 2015 we've also written a short paper outlining discussing some develop
...
-            /* ... */
-        }
-    }
-}
-```
+### Test/Parse Custom SQL Statements
+
+5. The build `example` executable can be used to test/parse specific custom SQL statements:
+     
+     Running, for example, `./example/example "SELECT * FROM test;"` produces:
```
Bouncner, line 28: `parse_simple_statement` as the binary name? Any ideas?

README.md
```diff
@@ -13,37 +13,58 @@ In March 2015 we've also written a short paper outlining discussing some develop
...
+	        const hsql::SQLStatement* statement = result.getStatement(0);
+	
+	        if (statement->isType(hsql::kStmtSelect)) {
+	            const auto* select = static_cast<const hsql::SelectStatement*>(statement);
+	            /* ... */
+	        }
+	    }
+	}
+	```
+
```
Bouncner, line 67: Last line in README: `HYRISE` > `Hyrise`

Bouncner (commented):
Just a few questions and suggestions. I think the proposed structure makes sense and is much better to read.

Bouncner:
Ping @klauck . :)

PR #245, opened 2024-06-26
Include cstdint for GCC-13

toge:
There are  several compilation errors with gcc 13.
You need to include cstdint and add the std:: prefix as described in the error message.

This fix is within the scope of the C++ standard and should not affect other compilers.

```
c++ -std=c++17 -O3 -c -o src/util/sqlhelper.o src/util/sqlhelper.cpp
In file included from src/sql/CreateStatement.h:4,
                 from src/sql/CreateStatement.cpp:1:
src/sql/ColumnType.h:29:34: error: 'int64_t' has not been declared
   29 |   ColumnType(DataType data_type, int64_t length = 0, int64_t precision = 0, int64_t scale = 0);
      |                                  ^~~~~~~
src/sql/ColumnType.h:29:54: error: 'int64_t' has not been declared
   29 |   ColumnType(DataType data_type, int64_t length = 0, int64_t precision = 0, int64_t scale = 0);
      |                                                      ^~~~~~~
src/sql/ColumnType.h:29:77: error: 'int64_t' has not been declared
   29 |   ColumnType(DataType data_type, int64_t length = 0, int64_t precision = 0, int64_t scale = 0);
      |                                                                             ^~~~~~~
src/sql/ColumnType.h:31:3: error: 'int64_t' does not name a type
   31 |   int64_t length;     // Used for, e.g., VARCHAR(10)
      |   ^~~~~~~
src/sql/ColumnType.h:5:1: note: 'int64_t' is defined in header '<cstdint>'; this is probably fixable by adding '#include <cstdint>'
    4 | #include <ostream>
  +++ |+#include <cstdint>
    5 | 
src/sql/ColumnType.h:32:3: error: 'int64_t' does not name a type
   32 |   int64_t precision;  // Used for, e.g., DECIMAL (6, 4) or TIME (5)
      |   ^~~~~~~
src/sql/ColumnType.h:32:3: note: 'int64_t' is defined in header '<cstdint>'; this is probably fixable by adding '#include <cstdint>'
src/sql/ColumnType.h:33:3: error: 'int64_t' does not name a type
   33 |   int64_t scale;      // Used for DECIMAL (6, 4)
      |   ^~~~~~~
src/sql/ColumnType.h:33:3: note: 'int64_t' is defined in header '<cstdint>'; this is probably fixable by adding '#include <cstdint>'
```


Bouncner:
Thanks for the fix. My only issue with this PR is that I am not a fan of prefixing simple basic types with `std::`. I would work without the prefix, wouldn't it?

Bouncner:
@toge : can you check if changing g++/gcc13 to version 14 in  the GH workflow already works? I think Ubuntu 24.04 ships GCC 14 or at least has install options for it.

src/util/sqlhelper.cpp
```diff
@@ -15,7 +16,7 @@ std::ostream& operator<<(std::ostream& os, const DatetimeField& datetime);
 std::ostream& operator<<(std::ostream& os, const FrameBound& frame_bound);
 
 std::string indent(uintmax_t num_indent) { return std::string(num_indent, '\t'); }
-void inprint(int64_t val, uintmax_t num_indent) { std::cout << indent(num_indent).c_str() << val << "  " << std::endl; }
+void inprint(std::int64_t val, uintmax_t num_indent) { std::cout << indent(num_indent).c_str() << val << "  " << std::endl; }
```
Bouncner, line 19: This is an example why I don't like the prefix. We would also need to add it to many other things (here uintmax_t) to make it consistent, right?

toge:
Sorry for not responding to your reply.
I haven't had a lot of time recently, but I will try to respond to the points you pointed out within the next couple of weeks.

toge:
@Bouncner 
Sorry for the delay.
I changed this PR to include `stdint.h` instead of `cstdint` to avoid compilation errors with gcc13 and later without `std::` prefix.

This modification will allow compiling with gcc13 without using `std::` and without affecting the previous gcc.

Could you review it again?

> Can you check if changing g++/gcc13 to version 14 in the GH workflow already works? I think Ubuntu 24.04 ships GCC 14 or at least has install options for it.

I checked sql-parser works fine on gcc14 without this PR on Fedora Linux 40.
This PR seems to be only for gcc 13 now.

src/sql/Expr.cpp
```diff
@@ -1,6 +1,7 @@
 #include "Expr.h"
 #include <stdio.h>
 #include <string.h>
+#include <stdint.h>
```
dey4ss, line 4: Pick here: Please have a look at the include order. Running the `format.sh`script should already fix this.

I also started a workflow run. Not sure what happened to the gcc-6 stage though. However, we already have a gcc-13 runner there since #242.

Bouncner: We might have a problem with Ubuntu 18. But where does node come from? It's from the GitHub action checkout? Curious how that can fail.

https://stackoverflow.com/a/74912791

dey4ss: Fixed with #246 
@toge to pass the CI, you need to merge the current master.

toge:
@Bouncner 
Sorry for suspending this PR.

Please go a head.
I am happy if this library works on much more environment.

Bouncner:
> Hi @toge, I would go ahead and merge the main branch, but since you are working on a fork, I am not able to.
> 
> Are you still interested in getting this PR landed? If you don't have the time, I would take it from here.

Oh ... did not know I can resolve conflicts on my own here.

Bouncner (approved):
LGTM.

PR #246, opened 2024-07-22
Fix CI build for Ubuntu 18.04

dey4ss:
In #245, we noticed that the Github checkout action now uses a Node.js version that is incompatible with our gcc-6/Ubuntu-18.04 stage. Thus, we pull the code manually for this stage.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,7 +59,7 @@ jobs:
           apt-get update
           apt-get install -y git
           git config --global --add safe.directory '*'
-          git clone https://github.com/hyrise/sql-parser .
+          git clone "${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}" .
           echo "checkout ${GITHUB_HEAD_REF}"
           git checkout $GITHUB_HEAD_REF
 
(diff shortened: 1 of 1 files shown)


Bouncner:
If it is too much of a pain, maybe upgrade to something more recent than GCC 6 and add a note?


Also: if checkout's node is too crappy to work on Ubuntu 18, might it make sense to checkout on our own?

dey4ss:
> Also: if checkout's node is too crappy to work on Ubuntu 18, might it make sense to checkout on our own?

I was afraid that the Github actions required node.js also for some more magical things during the runner image orchestration, but it seems to work just like that

.github/workflows/ci.yml
```diff
@@ -47,7 +47,21 @@ jobs:
...
+      - name: Checkout (Ubuntu 18.04)
+        if: matrix.name == 'gcc-6'
+        # Recent versions of Github's checkout action do not run on older Ubuntu versions because they use a too recent
+        # Node.js version. Thus, we have to checkout the code manually.
+        run: |
+          apt-get update
+          apt-get install -y git
+          git config --global --add safe.directory '*'
+          git clone "${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}" .
+          echo "checkout ${GITHUB_HEAD_REF}"
```
Bouncner, line 63: Remove?

Bouncner (approved):

PR #247, opened 2024-08-08
Make `COPY` statement options compliant to Postgres

dey4ss:
While fiddling around with a Postgres instance, I noticed that our syntax for `COPY` statements is not 100% compliant to Postgres. In detail, we used `COPY table_a FROM 'path/to/tbl' WITH FORMAT CSV;`, whereas Postgres uses `COPY table_a FROM 'path/to/tbl' WITH (FORMAT CSV);` – there are brackets around the options. Furthermore, the `WITH` keyword is optional [1].

The incorrect syntax was introduced in #139 (my bad 😕) and it slipped through the review back then. This PR makes the `COPY` statement compliant to Postgres again. Furthermore, I used the opportunity to add the encoding as another option. This allows us to apply any specified encoding for imported and exported tables in Hyrise.

[1] https://www.postgresql.org/docs/16/sql-copy.html


src/parser/bison_parser.y
```diff
@@ -438,20 +441,25 @@ execute_statement : EXECUTE IDENTIFIER {
 /******************************
  * Import Statement
  * IMPORT FROM TBL FILE 'test/students.tbl' INTO students
- * COPY students FROM 'test/students.tbl' [WITH FORMAT TBL]
+ * COPY students FROM 'test/students.tbl' WITH (FORMAT TBL, ENCODING 'Dictionary')
```
Bouncner, line 444: Do I get that right that the `WITH ...` part is now required?

dey4ss: In the end, this then becomes sth like: 
```SQL
COPY students FROM 'test/students.tbl' [ [WITH] ( [FORMAT TBL] [, ENCODING 'Dictionary'] ) ]
```
and is more the grammer than an example (not to talk about the different options here and that the options' order can be changed). I was not sure what to do there tbh. We could as well just ad a third example with format and encoding. We will also deliberately use the encoding option totally different from what it was meant to be, but that's not to matter here.

Have you seen the change to the  `AlterAction` constructor? I noticed we do not care about shadowing in the parser at all. Looking back at why we changed it in Hyrise (hyrise/hyrise#1877) we should not have issues in the parser, but what's your opinion on applying Hyrise's style here as well?

Bouncner: Any style improvements are welcome, sure. Can this be fixed automatically? Otherwise I am not sure if it is worth the effort.

Bouncner: And your call on the example. Maybe a third sounds good, before we have too much grammar-stuff in the examples.

dey4ss: It does not seem to be automatic, at least it does not seem to happened automatically in hyrise/hyrise#2040. With `-Weverything`, we have 35 occurrences of that. It's a rather rase case when this causes problems and I am a bit reluctant when it comes to adding errors looking at how many forks there are. So maybe leaving it as is is fine.

While playing arounf with it, I noticed that actually `-Wall -Werror` wad turned off for two years due to a typo in the flag name (and yes, again 100% my bad :/). I'll sneak in a fix.

Also added the extra example.

Makefile
```diff
@@ -42,7 +42,7 @@ relaxed_build ?= "off"
 ifeq ($(relaxed_build), on)
     $(warning $(NAME) will be built with most compiler warnings deactivated. This is fine if you want to test $(NAME) but will become an issue when you want to contribute code.)
 else
-    LIB_CLFAGS += -Wall -Werror
+    LIB_CFLAGS += -Wall -Werror
```
Bouncner, line 45: 👯

Makefile
```diff
@@ -69,6 +69,9 @@ $(LIB_BUILD): $(LIB_OBJ)
 $(SRCPARSER)/flex_lexer.o: $(SRCPARSER)/flex_lexer.cpp $(SRCPARSER)/bison_parser.cpp
 	$(CXX) $(LIB_CFLAGS) -c -o $@ $< -Wno-sign-compare -Wno-unneeded-internal-declaration -Wno-register
 
+$(SRCPARSER)/bison_parser.o: $(SRCPARSER)/bison_parser.cpp
+	$(CXX) $(LIB_CFLAGS) -c -o $@ $< -Wno-unused-but-set-variable
```
Bouncner, line 73: Short comment?

src/sql/ImportExportOptions.h
```diff
@@ -0,0 +1,24 @@
+#ifndef SQLPARSER_IMPORT_EXPORT_OPTIONS_H
+#define SQLPARSER_IMPORT_EXPORT_OPTIONS_H
+
+namespace hsql {
+
+// Name unchanged for compatibility.
```
Bouncner, line 6: I don't get this comment.

Bouncner (approved):
Only tiny things.

src/sql/ImportExportOptions.h
dey4ss: Dates back to when we added the `COPY` statement and, with it, the ability to export tables. Even though we use these file types for import and export, we kept the name so people with forks don't have to rename stuff in their code.

dey4ss: This thinking is also the reson why I added the encoding as another member of the statement rather than just passing the options object with both file type and encoding.

Makefile
dey4ss:
```c++
// Name unchanged for compatibility. Historically, this was only used for import statements before we introduced export
// statements (`COPY ... TO`). We did not change the enum name to accomodate forks.
```

Bouncner: Wrong place but okay. 😀

dey4ss: Lies nochmal 🚬

src/sql/ImportExportOptions.h
Bouncner: What name would you have chosen?

Makefile
dey4ss: Oh 😅

src/sql/ImportExportOptions.h
dey4ss: For the enum? In the grammar, we simply call it file type (so `FileType` as enum name):
https://github.com/hyrise/sql-parser/blob/9a67d2470ba053c9f9c63ebca7c100e6f73f0d2c/src/parser/bison_parser.y#L457
I remember that it was a big discussion back then and some insisted on compatibility. Anyway, I think it's kinda late for that now.

Bouncner: Can you briefly mention that we call it FileyType in the grammar, but won't adapt it here?
Not stating what the alternative is (nobody is aware of the probably old discussion) does not give away why the current name might not be perfect. :)

Bouncner: Feel free to zünd afterwards.

PR #250, opened 2024-09-10
[CI Test] Test new manual cloning

dey4ss:
Just a test for #249


dey4ss:
Close as #249 is merged.

PR #253, opened 2025-03-18
Update compiler versions in CI workflow

Bouncner:
Update the CI workflow to use the most recent GCC and clang versions available on Ubuntu 24.04.


Bouncner:
We need to remove the two outdated checks in the branch settings, before this could be merged.
Wanted to first get an approval, before I proceed.

dey4ss (approved):

PR #255, opened 2025-06-02
INTERVAL support, date pseudo literals TIMESTAMP and TIME literals.

kchasapis:
The interval was not allowed to be a column data type. With this patch we add this support.

We add support for the pseudo literals
CURRENT_TIME, CURRENT_TIMESTAMP, CURRENT_DATE.

Also, we add support for TIMESTAMP and TIME literals. The user can now explicitly set the type of the string value it follows with TIMESTAMP 'value' and TIME 'value.

DATE was already supported.


src/sql/CreateStatement.h
```diff
@@ -61,6 +61,7 @@ struct ColumnDefinition : TableElement {
   ColumnType type;
   bool nullable;
   std::vector<ReferencesSpecification*>* references;
+  Expr* defaultValue{nullptr};
```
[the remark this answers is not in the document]
kchasapis: @dey4ss I have deleted it.

src/sql/Expr.h
```diff
@@ -21,6 +21,11 @@ enum ExprType {
   kExprLiteralNull,
   kExprLiteralDate,
   kExprLiteralInterval,
+  kExprLiteralTimestamp,
+  kExprLiteralTime,
+  kExprCurrentTimestamp,
+  kExprCurrentDate,
+  kExprCurrentTime,
```
dey4ss, line 28: We discussed this internally and we prefer to have only an expression type per date/time/timestamp, not for the "current" version of the literal. Instead, please add a (boolean) flag to the `Expr` struct indicating whether the literal has the special "current" value.

Mini: The expression types are not really ordered by name. Could you bring them in the order date, time, timestamp, interval?

src/sql/Expr.h
```diff
@@ -188,8 +193,18 @@ struct Expr {
...
+
+  static Expr* makeTimeLiteral(char* val);
+
   static Expr* makeIntervalLiteral(int64_t duration, DatetimeField unit);
 
+  static Expr* makeCurrentTimestamp();
+
+  static Expr* makeCurrentDate();
+
+  static Expr* makeCurrentTime();
```
dey4ss, line 206: Please see my previous comment. The "current" versions of these functions are not required when we always pass a `char*`. If the pointer is a nullptr, we could automatically set the flag to mark the "current" literal version.

test/queries/queries-good.sql
```diff
@@ -1,4 +1,4 @@
-# This file contains a list of strings that are NOT valid SQL queries.
+# This file contains a list of strings that are valid SQL queries.
```
dey4ss, line 1: Nice catch!

src/parser/bison_parser.y
```diff
@@ -1176,19 +1188,22 @@ interval_literal : INTVAL duration_field { $$ = Expr::makeIntervalLiteral($1, $2
   free($2);
   $$ = Expr::makeIntervalLiteral(duration, $3);
 }
+| INTERVAL INTVAL datetime_field {
+  $$ = Expr::makeIntervalLiteral($2, $3);
+}
```
dey4ss, line 1193: Do you have a use case for this version? According to the SQL:2023 standard, only 
```
INTERVAL 'INTVAL' duration_field
```
would be valid (note the quotation marks). However, we supported some more versions according to Postgres. `INTERVAL INTVAL datetime_field` does not seem to work in Postgres (see [DB Fiddle](https://www.db-fiddle.com/f/bo7ZGJ9McxSjUESbWUfn58/0), statements in comments raise errors).

src/parser/bison_parser.y
```diff
@@ -1214,6 +1229,12 @@ param_expr : '?' {
   yyloc.param_list.push_back($$);
 };
 
+current_timestamp : CURRENT_TIMESTAMP { $$ = Expr::makeCurrentTimestamp(); };
+
+current_date : CURRENT_DATE { $$ = Expr::makeCurrentDate(); };
+
+current_time : CURRENT_TIME { $$ = Expr::makeCurrentTime(); };
```
dey4ss, line 1236: Can you please merge these with the regular versions? E.g., merge `current_timestamp` with `timestamp_literal` to the following:
```c++
timestamp_literal : TIMESTAMP STRING {
  $$ = Expr::makeTimestampLiteral($2);
}
| CURRENT_TIMESTAMP {
  $$ = Expr::makeTimestampLiteral(nullptr);
};
```

src/parser/flex_lexer.l
```diff
@@ -229,6 +229,9 @@ YEARS       TOKEN(YEARS)
 
 CURRENT[ \t\n]+ROW       TOKEN(CURRENT_ROW)
 CHARACTER[ \t\n]+VARYING TOKEN(CHARACTER_VARYING)
+CURRENT_TIMESTAMP        TOKEN(CURRENT_TIMESTAMP)
+CURRENT_DATE             TOKEN(CURRENT_DATE)
+CURRENT_TIME             TOKEN(CURRENT_TIME)
```
dey4ss, line 234: Can you please order lexicographically (date < time < timestamp)?

src/parser/bison_parser.y
```diff
@@ -265,6 +266,7 @@
 %type <expr>                   expr operand scalar_expr unary_expr binary_expr logic_expr exists_expr extract_expr cast_expr
 %type <expr>                   function_expr between_expr expr_alias param_expr
 %type <expr>                   column_name literal int_literal num_literal string_literal bool_literal date_literal interval_literal
+%type <expr>                   timestamp_literal time_literal current_timestamp current_date current_time
```
dey4ss, line 269: (see previous comment) Please remove the `current_` versions here.

src/sql/Expr.cpp
```diff
@@ -287,7 +314,8 @@ bool Expr::isType(ExprType exprType) const { return exprType == type; }
 
 bool Expr::isLiteral() const {
   return isType(kExprLiteralInt) || isType(kExprLiteralFloat) || isType(kExprLiteralString) || isType(kExprParameter) ||
-         isType(kExprLiteralNull) || isType(kExprLiteralDate) || isType(kExprLiteralInterval);
+         isType(kExprLiteralNull) || isType(kExprLiteralDate) || isType(kExprLiteralInterval) ||
+         isType(kExprLiteralTimestamp) || isType(kExprLiteralTime);
```
dey4ss, line 318: The "current" versions are missing here. However, we would suggest that they don't have own types anyways.

test/sql_tests.cpp
```diff
@@ -778,6 +778,219 @@ TEST(CastAsType) {
   ASSERT_EQ(stmt->selectList->front()->columnType.length, 8);
 }
 
+TEST(IntervalColumnTypeTest) {
+  auto result = SQLParserResult{};
+  SQLParser::parse(
+      "CREATE TABLE interval_table (c_interval INTERVAL )",
+      &result);
```
dey4ss, line 785: You can use `TEST_PARSE_SINGLE_SQL` here. This macro also asserts that the statement is valid and that it returns a single statement of the desired type and of the desired class.

test/sql_tests.cpp
```diff
@@ -778,6 +778,219 @@ TEST(CastAsType) {
...
+  ASSERT_EQ(stmt_year->values->at(0)->type, kExprLiteralInterval);
+  ASSERT_EQ(stmt_year->values->at(0)->ival, 1);
+  ASSERT_EQ(stmt_year->values->at(0)->datetimeField, kDatetimeYear);
+}
+
+TEST(CurrentDateTest) {
+  TEST_PARSE_SINGLE_SQL("SELECT * FROM students WHERE graduation_date = CURRENT_DATE;",
+                       kStmtSelect, SelectStatement, result, stmt);
+
+  ASSERT_TRUE(result.isValid());
```
dey4ss, line 817: pick here: `TEST_PARSE_SINGLE_SQL` already asserts this.

PR #256, opened 2025-07-28
CSV import options for DELIMITER, NULL, and QUOTE

Greenscreen23:
This branch implements more csv parsing options. The goal is that the following sql string is correctly parsed:
```
COPY "{}" FROM '{}' WITH (FORMAT CSV, DELIMITER '|', NULL '', QUOTE '"');
```

where delimiter is the delimiter of the csv file, null is the value of null values in the csv and quote is the string used to quote values.


Greenscreen23:
Should bison check that the values of delimiter, null, and quote are only set if the format is CSV? This check would have to be done when the options are set (if they are set after the format) and when the format is set (if the values are set before the format). Additionally, this check may have to be done in the import/export statement (if no format is set at all), but we could leave that out (if we don't want to infer the data type from the file string).

Greenscreen23:
> Some initial comments. I guess it makes sense to prohibit CSV options for binary files accordung to Postgres (https://www.postgresql.org/docs/current/sql-copy.html) at _some_ point (either in the parser or on the DBMS side), whatever is easier.

Sure, I've tried to implement something like that. Does this fit your needs?

test/queries/queries-bad.sql
```diff
@@ -53,6 +53,7 @@
 !SELECT * FROM t WHERE a = DATE '2000-01-01' + x DAYS;
 !SELECT * FROM t WHERE a = DATE '2000-01-01' + INTERVAL 'x' DAY;
 !SELECT * FROM t WHERE a = DATE '2000-01-01' + INTERVAL '3.3 DAYS';
+!COPY students FROM 'file_path' WITH (FORMAT TBL, DELIMITER '|', NULL '', QUOTE '"'); # Cannot have CSV options with non-CSV format
```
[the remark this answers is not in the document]
dey4ss: This is really necessary to ensure we clean up everything and do not leak when we call `YYERROR`. Please have a look at every path where you do that and add a case fot that, either in a full cpp test or here, so it will triggered when we do leak checking.

src/parser/bison_parser.y
```diff
@@ -517,7 +528,79 @@ import_export_options : import_export_options ',' FORMAT file_type {
...
+    if ($3->null) {
+      $1->csv_options->null = $3->null;
+      $3->null = nullptr;
+
+    }
+    if ($3->quote) {
+      $1->csv_options->quote = $3->quote;
+      $3->quote = nullptr;
+    }
+    delete $3;
```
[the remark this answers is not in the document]
Greenscreen23: I did something with a #define statement. I felt like having an assert like function makes more sense, but that doesn't work well with YYerror, as this resolves to a goto. So I had to use define.

src/parser/bison_parser.y
```diff
@@ -168,6 +168,7 @@
   hsql::RowLockWaitPolicy lock_wait_policy_t;
 
   hsql::ImportExportOptions* import_export_option_t;
+  std::pair<hsql::CsvOptionType, char*>* csv_option_t;
```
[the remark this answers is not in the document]
Greenscreen23: I just followed what was done for `ival_pair`. If you don't want it to be a pointer, I can change it :)

test/queries/queries-good.sql
```diff
@@ -74,6 +74,8 @@ COPY students FROM 'file_path' WITH (FORMAT TBL);
 COPY students FROM 'file_path' WITH (FORMAT CSV);
 COPY students FROM 'file_path' WITH (FORMAT BIN);
 COPY students FROM 'file_path' WITH (FORMAT BINARY);
+COPY students FROM 'file_path' WITH (FORMAT CSV, DELIMITER '|', NULL '', QUOTE '"');
+COPY students FROM 'file_path' WITH (DELIMITER '|', NULL '', FORMAT CSV, QUOTE '"');
```
Bouncner, line 78: Can you add "incomplete" queries as well? So leaving out NULL or QUOTE etc.

Same with auto format.

src/parser/bison_parser.y
```diff
@@ -517,7 +529,97 @@ import_export_options : import_export_options ',' FORMAT file_type {
 | ENCODING STRING {
   $$ = new ImportExportOptions{};
   $$->encoding = $2;
-};
+}
+| import_export_options ',' csv_option {
```
Bouncner, line 533: Given the current approach: can we use encoding and csv options?

test/queries/queries-good.sql
Greenscreen23: Sure, wil do.

src/parser/bison_parser.y
Greenscreen23: Yes, that should be parsed properly. I assume however, that for CSV files, we shouldn't have an encoding?

Bouncner: No, good that it works. The parser syntax just always confuses me. I need to take a pass once again.

Bouncner: Do we need to handle the code between 533 and 604 like that? A lot of repetition and I don't even know why. :D

src/parser/bison_parser.y
```diff
@@ -517,7 +529,97 @@ import_export_options : import_export_options ',' FORMAT file_type {
...
+    YYERROR;
+  }
+
+  if ($1->csv_options == nullptr) {
+    $1->csv_options = new CsvOptions{};
+  }
+
+  bool freed_ptr = false;
+  auto free_if_necessary = [&](char* ptr) {
+    if (ptr != nullptr) {
```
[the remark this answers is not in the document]
Bouncner: Ignore my comment. Wasn't aware of the way the parser requires deletes. @dey4ss cleared that up.

src/sql/statements.cpp
```diff
@@ -140,19 +140,50 @@ ExportStatement::ExportStatement(ImportType type)
...
+      break;
+    case CsvOptionType::Null:
+      if (null != nullptr) return false;
+      null = option->second;
+      break;
+    case CsvOptionType::Quote:
+      if (quote != nullptr) return false;
+      quote = option->second;
+      break;
+  }
```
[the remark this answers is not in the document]
dey4ss: It's really mixed here (even within this PR, that's why it caught my attention). I actually like the more implicit form, but I don't want to start a long discussion here. Braces after `if` are a must, though.

Bouncner (approved):
Zündable. But maybe still wait for @dey4ss to confirm. :)

dey4ss (approved):
🧨

PR #258, opened 2025-08-10
Add some newlines at the end of files

tjeyy:
`-Wnewline-eof` is only available in Clang, so missing linebreaks at the end of files are not noticed here.
But only because of the SQLParser is this warning manually disabled in Hyrise.
Let's change this.


Bouncner:
Can we add the check to the CI to ensure you got all newlines?

tjeyy:
Done.

Makefile
```diff
@@ -43,6 +43,9 @@ ifeq ($(relaxed_build), on)
     $(warning $(NAME) will be built with most compiler warnings deactivated. This is fine if you want to test $(NAME) but will become an issue when you want to contribute code.)
 else
     LIB_CFLAGS += -Wall -Werror
+    ifneq (,$(findstring clang,$(shell $(CXX) --version 2>/dev/null)))
+        LIB_CFLAGS += -Wnewline-eof
```
Bouncner, line 47: So the check is not part of Wall, yes?

Makefile
```diff
@@ -104,7 +107,7 @@ install:
 
 #######################################
 ############## Benchmark ##############
-#######################################
+########################################
```
Bouncner, line 110: Accident?

tjeyy: Whoops, yes!

Makefile
tjeyy: Of course!

Bouncner:
@dey4ss : anything to add from your side?

Issue #263, opened 2026-05-27
Feature request: support PostgreSQL `$N` and SQLite `:name` placeholder syntaxes


aleks-f:
### Background

The parser currently recognises a single placeholder form, `?` (positional, sequential). Most SQL dialects that use this parser as a frontend also accept one or both of these in production queries:

- **PostgreSQL** uses `$N` with N a positive integer — `WHERE a = $1 AND b = $2`. Order of appearance does not determine binding; the explicit number does.
- **SQLite** supports `:name` — `WHERE a = :user AND b = :id`. Bound by name.

Anyone embedding `hyrise/sql-parser` into a real database driver currently has to either pre-translate queries before parsing or work around the limitation downstream. Adding native support inside the parser is small, isolated, and unblocks those use cases without changing existing `?` behaviour.

### Proposed grammar

Two additional rules in the lexer, two additional alternatives in `param_expr`:

```
# flex_lexer.l (before the punctuation catch-all)
\$[0-9]+                 -> DOLLAR_PARAM   (ival = atoll(yytext + 1))
:[A-Za-z][A-Za-z0-9_]*   -> NAMED_PARAM    (sval = strdup(yytext + 1))
```

```
# bison_parser.y
param_expr
  : '?'           { /* existing */ }
  | DOLLAR_PARAM  { if ($1 < 1) YYERROR;            // reject $0
                    $$ = Expr::makeDollarParameter($1); ... }
  | NAMED_PARAM   { $$ = Expr::makeNamedParameter($1); ... }
  ;
```

The identifier pattern matches the existing `SQL_IDENTIFIER` rule so `:foo`, `:abc_123` parse and `:_x` does not — same surface as everywhere else in the grammar.

### Conflict analysis

- `$` is not currently a lexer token. No collision.
- `:` is in the punctuation character class but never referenced by any grammar rule (zero matches for `':'` in `bison_parser.y`); the standalone `:` token remains defined but unreachable. There is no `::` cast syntax in the grammar, so `:identifier` does not interfere with anything.
- `bison -v` reports no new shift/reduce or reduce/reduce conflicts on the resulting grammar (verified locally).

### Proposed AST

Three `ExprType` values rather than one with a discriminator field:

```
enum ExprType {
  ...
  kExprParameter,         // ?
  kExprParameterDollar,   // $N — ival holds N (1-based, preserves user intent)
  kExprParameterNamed,    // :name — name holds the identifier
  ...
};

[...]

PR #264, opened 2026-05-27
Add PostgreSQL $N and SQLite :name placeholder syntaxes (#263)

aleks-f:
Extends the parser to recognise two placeholder forms alongside the existing ?:

- $N (PostgreSQL 1-based explicit numbering, e.g. WHERE a = $1)
- :name (SQLite named, e.g. WHERE a = :user)

The ? path is unchanged at every layer (lexer, parser, AST, result ordering); existing tests under test/prepare_tests.cpp pass untouched.

Implementation:

- New lexer rules for \$[0-9]+ and :[A-Za-z][A-Za-z0-9_]* placed before the punctuation catch-all. $ was previously unused; : was in the punctuation class but never referenced by any grammar rule, so the longer :identifier match wins without conflict.
- New bison tokens DOLLAR_PARAM (carries the int N) and NAMED_PARAM (carries the identifier). param_expr now has three alternatives. $0 is rejected via YYERROR at parse time.
- Three ExprType values appended (no ordinal change): kExprParameter for ?, kExprParameterDollar for $N (ival = N, 1-based), and kExprParameterNamed for :name (name = identifier). Distinct enum values rather than one with discriminator fields keep consumer switch statements explicit and round-trip printing straightforward.
- Factory functions Expr::makeDollarParameter and makeNamedParameter. isLiteral returns true for all three. The destructor already frees name so no further changes are needed.
- The top-level input rule's renumber loop only touches kExprParameter; $N keeps its explicit ival, :name keeps its name. SQLParserResult::addParameter still sorts by ival, which now orders $N by explicit position and clusters named placeholders.
- sqlhelper print case extended to handle the two new ExprType values.

Tests:

- Three new cases in test/prepare_tests.cpp covering $N, $N declared out of order (verifying ival-sorted result vector), and :name.
- Three new good queries and two new bad queries ($0, lone $) in test/queries/.
- bison -v reports no new shift/reduce or reduce/reduce conflicts.
- make test passes all three checks (SQL tests, valgrind, grammar).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,9 +72,12 @@ $(LIB_BUILD): $(LIB_OBJ)
 	$(LIBLINKER) $(LIB_LFLAGS) $(LIB_BUILD) $(LIB_OBJ)
 
 # The auto-generated code from bison and flex contains some parts the compiler complains about with -Wall.
+# bison_parser.cpp #includes flex_lexer.h, so bison_parser.o must wait for flex_lexer.cpp regeneration
+# to finish (which also produces flex_lexer.h). Without this dep, parallel make races and bison_parser.o
+# can start before hsql_lex is declared, producing 'hsql_lex was not declared' errors.
 $(SRCPARSER)/flex_lexer.o: $(SRCPARSER)/flex_lexer.cpp $(SRCPARSER)/bison_parser.cpp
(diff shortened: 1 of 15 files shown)


aleks-f:
I just noticed bison upgrade also went in. If you want that reverted, let me know.

PR #265, opened 2026-06-11
Add support for three-part column references

malensek:
This adds support for three-part column references:

```sql
SELECT schema.table.column FROM table;
```
So something like `SELECT catalog.students.grade FROM students;` will populate the schema field in `Expr`:

```c++
expr->schema = "catalog";
expr->table = "students";
expr->name = "grade";
```

The changes are fairly light:
  - Added a grammar entry for `IDENTIFIER '.' IDENTIFIER '.' IDENTIFIER`
  - Added `Expr::makeColumnRef(char* schema, char* table, char* name)`
  - Added a test case covering this scenario

I have confirmed this passes the test suite and leak test on Linux 6.18.35-1-lts and test suite on macOS 26.5.1. I also see that the repo has the Bison output checked in so I did so as well for this PR, but please let me know if I should exclude it or need to run with the same version of Bison that was previously used to generate the files.

Thanks for this great library, it's very clean and easy to work with!

### DIFF
--- a/src/parser/bison_parser.y
+++ b/src/parser/bison_parser.y
@@ -1230,6 +1230,7 @@ between_expr : operand BETWEEN operand AND operand { $$ = Expr::makeBetween($1, 
 
 column_name : IDENTIFIER { $$ = Expr::makeColumnRef($1); }
 | IDENTIFIER '.' IDENTIFIER { $$ = Expr::makeColumnRef($1, $3); }
+| IDENTIFIER '.' IDENTIFIER '.' IDENTIFIER { $$ = Expr::makeColumnRef($1, $3, $5); }
 | '*' { $$ = Expr::makeStar(); }
 | IDENTIFIER '.' '*' { $$ = Expr::makeStar($1); };
 
--- a/src/parser/bison_parser.cpp
+++ b/src/parser/bison_parser.cpp
@@ -1,8 +1,9 @@
-/* A Bison parser, made by GNU Bison 3.0.4.  */
+/* A Bison parser, made by GNU Bison 3.8.2.  */
 
 /* Bison implementation for Yacc-like parsers in C
 
-   Copyright (C) 1984, 1989-1990, 2000-2015 Free Software Foundation, Inc.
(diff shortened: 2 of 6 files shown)


src/parser/bison_parser.y
```diff
@@ -1230,6 +1230,7 @@ between_expr : operand BETWEEN operand AND operand { $$ = Expr::makeBetween($1,
 
 column_name : IDENTIFIER { $$ = Expr::makeColumnRef($1); }
 | IDENTIFIER '.' IDENTIFIER { $$ = Expr::makeColumnRef($1, $3); }
+| IDENTIFIER '.' IDENTIFIER '.' IDENTIFIER { $$ = Expr::makeColumnRef($1, $3, $5); }
```
dey4ss: There already is a `table_name` definition. Can we just reuse it here, e.g., replace `IDENTIFIER '.' IDENTIFIER {...}` with `table_name '.' IDENTIFIER {...}` and use the already parsed schema and table name in the body?

src/parser/bison_parser.y
```diff
@@ -1230,6 +1230,7 @@ between_expr : operand BETWEEN operand AND operand { $$ = Expr::makeBetween($1,
...
 | '*' { $$ = Expr::makeStar(); }
 | IDENTIFIER '.' '*' { $$ = Expr::makeStar($1); };
```
dey4ss: For consistency: I guess this does also apply to `SELECT <table_name>.*`. Can you add it here as well (with a test)?

dey4ss (commented):
Thank you for your contribution! I have two small questions.

src/parser/bison_parser.y
malensek: Will do!

src/parser/bison_parser.y
malensek: Oh, that is a good idea... Initially I tried this directly with `table_name` but it produced a conflict between `table_name` and `column_name`, but perhaps a new definition could be used in both places. Let me know if there a better way to do it... I'll update the PR in a bit.

malensek:
Thanks a lot for reviewing this -- just updated the PR with the changes. Let me know if I can tweak this more.

